PR #52, opened 2020-06-29
Fix arch detection on windows arm64 devices

### LINKED ISSUE #51: [dotnet-install.ps1] Incorrect auto architecture when running from Powershell classic on Windows ARM64
When running `dotnet-install.ps1` from Powershell classic on Windows ARM64 without an `-Architecture` specified, the installer will resolve to downloading an `x86` build of the sdk. 

The reason seems to be [this line](https://github.com/dotnet/install-scripts/blob/ede04c0065c25d30ad05ae6820ef8260c55e976c/src/dotnet-install.ps1#L158), Powershell classic runs as emulated x86 so the return value is, accordingly, `x86`. As expected, the issue does not impact powershell core.

We should probably update the script so that the processor architecture is determined via different means. One potential solution is using the windows command `wmic OS get OSArchitecture`, which reports the correct OS architecture. If you agree with this solution I could perhaps try to contribute a PR that fixes the problem.

We've been seeing this issue in the windows arm64 builds of `dotnet/runtime`, which is using `dotnet-install.ps1` and powershell classic to download its bootstrapping sdk. cc @safern @ViktorHofer

eiriktsarpalis:
Fixes #51

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -154,8 +154,10 @@ function Invoke-With-Retry([ScriptBlock]$ScriptBlock, [int]$MaxAttempts = 3, [in
 function Get-Machine-Architecture() {
     Say-Invocation $MyInvocation
 
+    # emulated x86 processes (such as powershell classic on ARM64) can report the wrong architecture
+    # read the variable from the machine level to get the true OS architecture.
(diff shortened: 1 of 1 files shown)


bozturkMSFT:
Closing as duplicate

PR #54, opened 2020-07-10
Azure functions project, tests and publishing

bozturkMSFT:
Delivers the following issues:
- Integrate KUSTO client to Azure Function [#1121189](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1121189)
- Report SH download Metric events [#1121192](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1121192)
- Update CI\CD to support Azure Function [#1121189](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1121189) (partial delivery. Key vault is not accessible yet. Publishing is implemented, but disabled due to authorization issue).

List of changes in this PR:
- Created a C# projects to hold azure functions
- Some unit tests were written
- Kusto integration was implemented
- Configured the project so that `dotnet publish` pushes the functions to corresponding azure resource, provided that the key vault is accessible (Key vault is currently not available).


src/MonitoringFunctions/Functions/DownloadPs1.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Extensions.Logging;
+
+namespace MonitoringFunctions.Functions
+{
+    public static class DownloadPs1
```
donJoseLuis, line 9: very minor: Type names should not be "commands" or "actions".  Instead, please consider ```Ps1Downloader```

src/MonitoringFunctions/Functions/DownloadSh.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Extensions.Logging;
+
+namespace MonitoringFunctions.Functions
+{
+    public static class DownloadSh
```
donJoseLuis, line 9: minor: this is essentially the same logic as the previous class, the only variant is the URI

src/MonitoringFunctions/HelperMethods.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace MonitoringFunctions
+{
+    internal static class HelperMethods
+    {
+        /// <summary>
+        /// Single instance is sufficient for the whole app.
+        /// From official docs: HttpClient is intended to be instantiated once and re-used throughout the life of an application.
+        /// https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netcore-3.1#remarks
+        /// </summary>
+        private static readonly HttpClient _httpClient = new HttpClient();
```
donJoseLuis, line 20: this is IDisposable

src/MonitoringFunctions/HelperMethods.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        /// <summary>
+        /// Tests if the contents of the given url can be accessed from current environment.
+        /// Saves results to Kusto.
+        /// </summary>
+        /// <param name="log"><see cref="ILogger"/> that is used to report log information.</param>
+        /// <param name="monitorName">Name of this monitor to be included in the logs and in the data sent to Kusto.</param>
+        /// <param name="url">Url that this method will attempt to access.</param>
+        /// <returns></returns>
+        internal static async Task CheckAndReportUrlAccess(ILogger log, string monitorName, string url, IDataService dataService)
```
donJoseLuis, line 30: async method names should end on the suffix Async

donJoseLuis, line 30: consider adding an optional cancellation token last parameter to the signature

src/MonitoringFunctions/HelperMethods.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            HttpResponseMessage response = await _httpClient.GetAsync(url, HttpCompletionOption.ResponseHeadersRead).ConfigureAwait(false);
+
+            await dataService.ReportUrlAccess(monitorName, response);
```
donJoseLuis, line 34: configure the awaitable to false, this could cause a weird freeze in the host process

src/MonitoringFunctions/HelperMethods.cs
donJoseLuis, line 30: is this the class that performs the download of the scripts? if yes, should you test for the resources actually being downloaded to disk?

src/MonitoringFunctions/IDataService.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    internal interface IDataService : IDisposable
+    {
+        /// <summary>
+        /// Stores the details of the <see cref="HttpResponseMessage"/> in the underlying data store.
+        /// </summary>
+        /// <param name="monitorName">Name of the monitor that will be associated with this data.</param>
+        /// <param name="httpResponse">Http response data to be stored.</param>
+        /// <returns>A task, tracking the initiated async operation. Errors should be reported through exceptions.</returns>
+        Task ReportUrlAccess(string monitorName, HttpResponseMessage httpResponse);
```
donJoseLuis, line 17: if this is meant to be implemented as an async method, consider suffixing the method name with "Async" + cancellation token optional param

src/MonitoringFunctions/MonitoringFunctions.csproj
```diff
@@ -0,0 +1,26 @@
...
+  <ItemGroup>
+    <None Update="host.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="local.settings.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+    </None>
+  </ItemGroup>
+</Project>
```
donJoseLuis, line 26: white space

src/MonitoringFunctions/Properties/serviceDependencies.dotnet-install-daemon-functions - Zip Deploy.json
```diff
@@ -0,0 +1,16 @@
...
+      "secretStore": null
+    },
+    "storage1": {
+      "resourceId": "/subscriptions/[parameters('subscriptionId')]/resourceGroups/[parameters('resourceGroupName')]/providers/Microsoft.Storage/storageAccounts/monitoringfuncsstorage",
+      "type": "storage.azure",
+      "connectionId": "AzureWebJobsStorage",
+      "secretStore": "AzureKeyVault"
+    }
+  }
+}
```
donJoseLuis, line 16: white space

src/MonitoringFunctions/Properties/serviceDependencies.json
```diff
@@ -0,0 +1,12 @@
...
+    "secrets1": {
+      "type": "secrets",
+      "connectionId": "monitoring-functions-push-connection-string"
+    },
+    "storage1": {
+      "type": "storage",
+      "connectionId": "AzureWebJobsStorage"
+    }
+  }
+}
```
donJoseLuis, line 12: white space

src/MonitoringFunctions/Properties/serviceDependencies.local.json
```diff
@@ -0,0 +1,9 @@
+{
+  "dependencies": {
+    "storage1": {
+      "resourceId": "/subscriptions/[parameters('subscriptionId')]/resourceGroups/[parameters('resourceGroupName')]/providers/Microsoft.Storage/storageAccounts/monitoringfuncsstorage",
+      "type": "storage.azure",
+      "connectionId": "AzureWebJobsStorage"
+    }
+  }
+}
```
donJoseLuis, line 9: white space

src/MonitoringFunctions/Providers/DummyDataService.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace MonitoringFunctions.Providers
+{
+    internal class DummyDataService : IDataService
+    {
+        public async Task ReportUrlAccess(string monitorName, HttpResponseMessage httpResponse)
```
donJoseLuis, line 9: suffix "Async" to method name + cancellation token optional param

src/MonitoringFunctions/Providers/DummyDataService.cs
```diff
@@ -0,0 +1,19 @@
...
+        {
+            await Task.Delay(new Random().Next(200, 4000)).ConfigureAwait(false);
```
donJoseLuis, line 11: consider overload taking cancellation token

src/MonitoringFunctions/Providers/KustoDataService.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace MonitoringFunctions
+{
+    internal class KustoDataService : IDataService
+    {
+        private const string serviceNameAndRegion = "dotnetinstallcluster.eastus2";
```
donJoseLuis, line 16: constants names should start with a CAP

src/MonitoringFunctions/Providers/KustoDataService.cs
```diff
@@ -0,0 +1,104 @@
...
+using Kusto.Ingest;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace MonitoringFunctions
+{
+    internal class KustoDataService : IDataService
```
donJoseLuis, line 14: if you don't intend for this class to be a parent, consider sealing

src/MonitoringFunctions/Providers/KustoDataService.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        private const string serviceNameAndRegion = "dotnetinstallcluster.eastus2";
+        private const string databaseName = "dotnet_install_monitoring_database";
+        private const string tableName = "DaemonLogs";
+
+        private static IEnumerable<ColumnMapping> DaemonLogsJsonMapping { get; }
+
+        static KustoDataService()
```
donJoseLuis, line 22: nice work having ramped up so fast in KUSTO

src/MonitoringFunctions/Providers/KustoDataService.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            DaemonLogsJsonMapping = new List<ColumnMapping>()
```
donJoseLuis, line 24: very minor:  this could have been ```ColumnMapping[]```

src/MonitoringFunctions/Providers/KustoDataService.cs
```diff
@@ -0,0 +1,104 @@
...
+                IngestionMappingKind = Kusto.Data.Ingestion.IngestionMappingKind.Json,
+                IngestionMappings = DaemonLogsJsonMapping
+            };
+            ingestProps.Format = DataSourceFormat.json;
+
+            using MemoryStream memStream = new MemoryStream();
+            using StreamWriter writer = new StreamWriter(memStream);
+
+            writer.WriteLine(
+                @"{{ ""monitor_name"":""{0}"", ""timestamp"":""{1}"", " +
```
donJoseLuis, line 88: consider abstracting the "events" in a structure

src/MonitoringFunctions/Providers/KustoDataService.cs
```diff
@@ -0,0 +1,104 @@
...
+                @"""requested_url"":""{2}"", ""http_response_code"":""{3}"" }}",
+                monitorName, DateTime.UtcNow,
+                httpResponse.RequestMessage.RequestUri, (int)httpResponse.StatusCode);
+
+            writer.Flush();
+            memStream.Seek(0, SeekOrigin.Begin);
+
+            await ingestClient.IngestFromStreamAsync(memStream, ingestProps, leaveOpen: true).ConfigureAwait(false);
```
donJoseLuis, line 96: if there is an overload that accepts a cancellation token, we should instead use that

tests/MonitoringFunctions.Test/TestHelperMethods.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    [TestClass]
+    public class TestHelperMethods
+    {
+        [TestMethod]
+        public async Task TestCheckUrlAccess()
+        {
+            using IDataService dataService = new DummyDataService();
+            await HelperMethods.CheckAndReportUrlAccess(NullLogger.Instance, "test_run", "https://www.microsoft.com", dataService);
+        }
```
donJoseLuis, line 18: what is being tested?

src/MonitoringFunctions/IDataService.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace MonitoringFunctions
+{
+    internal interface IDataService : IDisposable
```
ostorc, line 9: Why is IDisposable used here? It is for possible future implementations?

tests/MonitoringFunctions.Test/TestHelperMethods.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        [DataRow("definitely not a url")]
+        [DataRow("https://www.somewebsitethatdontexist1239420312392304.com")]
```
ostorc, line 21: This URL may fail this test one time (highly unlikely, but still). Maybe use instead some reserved URL (like 0.com)

src/MonitoringFunctions/Functions/DownloadSh.cs
bozturkMSFT: What do you suggest we do here?

src/MonitoringFunctions/IDataService.cs
bozturkMSFT: Yes. I find it likely that a data service we will implement in the future will have some long living object that needs to be disposed. If I don't include this in the interface today, it will be harder to update the code to call `Dispose` method at correct places. I may be wrong and this may never be used. I just chose the option that looks safer.

src/MonitoringFunctions/HelperMethods.cs
bozturkMSFT: In this case, we need it until the app exits and it will be automatically disposed when that happens.

src/MonitoringFunctions/HelperMethods.cs
bozturkMSFT: This checks if given url is accessible (http status code is 200). I will use this to also check if sdk bits are available at the blob storage, so this method shouldn't be concern about the content at the url.

src/MonitoringFunctions/MonitoringFunctions.csproj
bozturkMSFT: Must be placed by github. There is no whitespace in the file.

src/MonitoringFunctions/Properties/serviceDependencies.dotnet-install-daemon-functions - Zip Deploy.json
bozturkMSFT: Must be placed by github. There is no whitespace in the file.

src/MonitoringFunctions/Properties/serviceDependencies.json
bozturkMSFT: Must be placed by github. There is no whitespace in the file.

src/MonitoringFunctions/Properties/serviceDependencies.local.json
bozturkMSFT: Must be placed by github. There is no whitespace in the file. Besides, these files are managed by visual studio. Why should we concern ourselves with white spaces?

src/MonitoringFunctions/Providers/KustoDataService.cs
bozturkMSFT: Unfortunately, there is not.

tests/MonitoringFunctions.Test/TestHelperMethods.cs
bozturkMSFT: This tests that our method can successfully make an HTTP request and get result 200.

tests/MonitoringFunctions.Test/TestHelperMethods.cs
bozturkMSFT: Thanks, I didn't know about 0.com. Is it a reserved URL or a an invalid one? Can domain names begin with a number?

ostorc: The specification doesn't says anything about that (https://tools.ietf.org/html/rfc3986#section-3.2.2). When I looked up the domain in the [whois tool](https://lookup.icann.org/lookup) it said that it is registered under IANA.

bozturkMSFT: Awesome!

src/MonitoringFunctions/Functions/DownloadSh.cs
donJoseLuis: consider consolidating the logic in a single place, then leveraging it with the parameter sets for ps1 & sh, respectively.  Moving forward, enhancements & bug fixes happen in a single place

src/MonitoringFunctions/HelperMethods.cs
donJoseLuis: you can make that same argument for most classes.  This being a static class, consider a ```CleanUp``` type method that essentially cleans up such resources & invalidates the instance.  Moving forward, callers will have the option to reclaim those resources during the life of the process.

tests/MonitoringFunctions.Test/TestHelperMethods.cs
donJoseLuis: my question relates to the absence of assertions in the test case method.  The scheme is: if the method completes, the method passes, without verifications.

src/MonitoringFunctions/Functions/DownloadSh.cs
bozturkMSFT: I think that is what I did when I created a helper method `CheckAndReportUrlAccessAsync` and called it from both functions. That is where we will do the fixes and improvements, and both functions will be benefiting. Functions are called by azure, so I don't have the option to pass additional parameters.

donJoseLuis: Sounds like you've already thought about this & have implemented the best possible solution.  I am happy.  Let's close on this one.

tests/MonitoringFunctions.Test/TestHelperMethods.cs
bozturkMSFT: `Assert.ThrowsException` doesn't seem applicable here. Method assumes that the test will succeed when there is an exception of given type, whereas in our case an exception means failure. As we discussed, instead of adding Assert for the sake of Assert, I added a comment with hopes of clarifying the situation a bit.

bozturkMSFT:
Force pushing seems to have confused Github. My changes are not visible here in the changes tab. Closing this to open a new one.

PR #58, opened 2020-07-17
Allow Bypass list to be passed into the proxy

ZEisinger:
Required for actions/setup-dotnet#71. Agents can set proxy settings for their runs which include addresses which should not go through the proxy. This script doesn't allow that value to be specified for windows. In the shell script this is handled by wget and curl both honoring the environment variables `https_proxy` and `no_proxy`.

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -69,6 +69,8 @@
 .PARAMETER ProxyUseDefaultCredentials
     Default: false
     Use default credentials, when using proxy address.
+.PARAMETER ProxyBypassList
+    If set with ProxyAddress, will provide the list of comma separated urls that will bypass the proxy
 .PARAMETER SkipNonVersionedFiles
     Default: false
     Skips installing non-versioned files if they already exist, such as dotnet.exe.
@@ -96,6 +98,7 @@ param(
    [string]$FeedCredential,
    [string]$ProxyAddress,
    [switch]$ProxyUseDefaultCredentials,
+   [string]$ProxyBypassList="",
    [switch]$SkipNonVersionedFiles,
    [switch]$NoCdn
 )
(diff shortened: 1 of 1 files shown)


bozturkMSFT:
Hi @ZEisinger ,
 Can you give me a bit more info on your use cases? For instance, how do you acquire the script? Do you have a local copy that you retain or are you downloading it from www.dot.net/v1/dotnet-install.ps1 before each run? Are you blocked by this PR?
 Unfortunately, we currently don't have a strong testing mechanism to ensure that the changes to the scripts won't break any existing scenarios. Therefore, we refrain from making changes.

src/dotnet-install.ps1
```diff
@@ -253,7 +256,11 @@ function GetHTTPResponse([Uri] $Uri)
 
             if($ProxyAddress) {
                 $HttpClientHandler = New-Object System.Net.Http.HttpClientHandler
-                $HttpClientHandler.Proxy =  New-Object System.Net.WebProxy -Property @{Address=$ProxyAddress;UseDefaultCredentials=$ProxyUseDefaultCredentials}
+                $HttpClientHandler.Proxy =  New-Object System.Net.WebProxy -Property @{
+                    Address=$ProxyAddress;
+                    UseDefaultCredentials=$ProxyUseDefaultCredentials;
+                    BypassList = $ProxyBypassList.Split(",")
```
nohwnd, line 262: I would make this array of strings instead of splitting on comma.

src/dotnet-install.ps1
```diff
@@ -96,6 +98,7 @@ param(
    [string]$FeedCredential,
    [string]$ProxyAddress,
    [switch]$ProxyUseDefaultCredentials,
+   [string]$ProxyBypassList="",
```
nohwnd, line 101: Defaulting to empty string is not very safe because you can always pass explicit null when the parameter is not mandatory.

ZEisinger:
@bozturkMSFT Actions follow a consistent pattern for [declaring and consuming proxy settings](https://docs.github.com/en/actions/hosting-your-own-runners/using-a-proxy-server-with-self-hosted-runners) which includes the ability to specify when to skip the proxy (`no_proxy` environment variable or `BypassList` in dotnet). The windows install script has the ability to specify the proxy address to use, but there no way of specifying when to skip the proxy.

The action currently consumes these scripts from a local copy, but checks the feed for updates in CI tests. I know there is typically some lead time on changes to these script so I opened the PR here before my PR is ready in setup-dotnet. If needed, I can disable the CI test temporarily until this change is available at www.dot.net.

src/dotnet-install.ps1
ZEisinger: Thanks @nohwnd, I have updated with the suggested changes.

src/dotnet-install.ps1
```diff
@@ -69,6 +69,8 @@
 .PARAMETER ProxyUseDefaultCredentials
     Default: false
     Use default credentials, when using proxy address.
+.PARAMETER ProxyBypassList
+    If set with ProxyAddress, will provide the list of comma separated urls that will bypass the proxy
```
bozturkMSFT, line 73: Can we update this to reflect the latest changes? Instead of comma separated urls, we have a string array now.

nohwnd: Imho this is correct for both cases "abc", "abc" "abc, abc" , both could be considered comma separated list, but just the first is an array of strings.

PR #59, opened 2020-07-21
Added DryRun Functions and DataServiceFactory

bozturkMSFT:
Delivers the following ADO work items:
- [#1121193](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1121193) Extend Azure Function to download LTS
- [#1121196](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1121196) Extend Azure Function to download 3.1 channel latest 
- [#1121197](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1121197) Extend Azure Function to download 3.0 runtime

The changes include:
- 3 new functions: one for each of the work items
- New types to streamline pushing data to kusto using JSON.
- New types and a helper method to run install scripts in DryRun mode and get the executions results
- Unit tests


src/MonitoringFunctions/DataService/Kusto/KustoTable.cs
```diff
@@ -0,0 +1,57 @@
...
+                IngestionMapping = new IngestionMapping()
+                {
+                    IngestionMappingKind = global::Kusto.Data.Ingestion.IngestionMappingKind.Json,
+                    IngestionMappings = columnMappings
+                },
+                Format = DataSourceFormat.json
+            };
+        }
+
+        public async Task InsertRow(T row, CancellationToken cancellationToken = default)
```
ostorc, line 44: Consider using suffix Async

src/MonitoringFunctions/DataService/DataServiceFactory.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using MonitoringFunctions.Providers;
+using System;
+
+namespace MonitoringFunctions
+{
+    internal class DataServiceFactory
```
ostorc, line 8: Maybe make this class also sealed, since it has no virtual members?

ostorc: Also maybe make this class static? I have only seen the usage:
```c#
new DataServiceFactory().GetDataService()
```

src/MonitoringFunctions/DataService/Kusto/KustoTable.cs
bozturkMSFT: Nice catch. Thanks!

src/MonitoringFunctions/DataService/Kusto/DirectJsonMappingResolver.cs
```diff
@@ -0,0 +1,36 @@
...
+    internal sealed class DirectJsonMappingResolver : IJsonColumnMappingResolver
+    {
+        public IEnumerable<ColumnMapping> GetColumnMappings<T>() where T : IKustoTableRow
+        {
+            DefaultContractResolver contractResolver = new DefaultContractResolver();
+            var contract = contractResolver.ResolveContract(typeof(T)) as JsonObjectContract;
+
+            if (contract == null)
+            {
+                throw new Exception($"Failed to resolve contract. Automatic column mapping is not possible with this type {typeof(T)}.");
```
donJoseLuis, line 23: minor: is there a more relevant Exception type for this case?

src/MonitoringFunctions/DataService/Kusto/IKustoTableRow.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace MonitoringFunctions.DataService.Kusto
+{
+    /// <summary>
+    /// This interface marks model classes that can be inserted into a kusto table as a row.
+    /// </summary>
+    internal interface IKustoTableRow
+    {
```
donJoseLuis, line 9: minor: I'm not yet understanding this interface.  Perhaps I need a more thorough review.

src/MonitoringFunctions/DataService/DataServiceFactory.cs
bozturkMSFT: Thanks for the suggestions. I am not sure if I feel strongly about either of those. 
Class is already internal and safe to inherit from, even though there may be no reason to do so.
Making it static will save us an allocation, but it will make it harder if we needed instances to configure and pass around in the future.

src/MonitoringFunctions/DataService/Kusto/KustoTable.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            using IKustoQueuedIngestClient ingestClient = KustoIngestFactory.CreateQueuedIngestClient(_connectionStringBuilder);
+
+            string serializedData = JsonConvert.SerializeObject(row);
+            byte[] serializedBytes = Encoding.UTF8.GetBytes(serializedData);
+
+            using MemoryStream dataStream = new MemoryStream(serializedBytes);
+
+            // IKustoQueuedIngestClient doesn't support cancellation at the moment. Update the line below if it does in the future.
+            await ingestClient.IngestFromStreamAsync(dataStream, _ingestionProperties, leaveOpen: true);
```
donJoseLuis, line 54: minor:  in cases like this, where an instance is disposable, created with a  ```using``` and then passed to a method, sometimes results in an unexpected disposed instance in the called method's stack.
**Question** have you tested this behavior?  My solution tosuch cases has been removing the using.

src/MonitoringFunctions/DataService/Providers/DummyDataService.cs
```diff
@@ -14,9 +14,15 @@ public async Task ReportUrlAccessAsync(string monitorName, HttpResponseMessage h
...
+        public async Task ReportScriptExecutionAsync(string monitorName, string scriptName, string commandLineArgs, string error, CancellationToken cancellationToken = default)
+        {
+            await Task.Delay(new Random().Next(200, 4000), cancellationToken).ConfigureAwait(false);
+        }
+
         public void Dispose()
         {
             // Do nothing
         }
+
```
donJoseLuis, line 26: consider setting an instance boolean to true the first time dispose is called

src/MonitoringFunctions/DataService/Providers/DummyDataService.cs
```diff
@@ -14,9 +14,15 @@ public async Task ReportUrlAccessAsync(string monitorName, HttpResponseMessage h
             await Task.Delay(new Random().Next(200, 4000), cancellationToken).ConfigureAwait(false);
         }
 
+        public async Task ReportScriptExecutionAsync(string monitorName, string scriptName, string commandLineArgs, string error, CancellationToken cancellationToken = default)
```
donJoseLuis, line 17: minor: consider a style where after x-many parameters, each parameter will be in its' own line.  Team-wide, we need coding styles sorted out.

src/MonitoringFunctions/DataService/Providers/KustoDataService.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Threading.Tasks;
+
+namespace MonitoringFunctions
+{
+    internal sealed class KustoDataService : IDataService
+    {
+        private const string ServiceNameAndRegion = "dotnetinstallcluster.eastus2";
+        private const string DatabaseName = "dotnet_install_monitoring_database";
+
+        private readonly KustoTable<HttpRequestLogEntry> HttpRequestLogsTable;
```
donJoseLuis, line 18: minor: please consider prefixing instance fields with underscore; thus, ```_httpRequestLogsTable```

src/MonitoringFunctions/Functions/DryRunUrlChecker.cs
```diff
@@ -0,0 +1,126 @@
...
+            {
+                log.LogError($"Primary Url was not found for channel {additionalCmdArgs}");
+                await dataService.ReportScriptExecutionAsync(monitorName, scriptName, commandLineArgs,
+                    "Failed to parse primary url from the following DryRun execution output: " + results.Output
+                    , cancellationToken).ConfigureAwait(false);
+                return;
+            }
+
+            // Validate URL accessibility
+            await HelperMethods.CheckAndReportUrlAccessAsync(log, monitorName, dryRunResults.PrimaryUrl, dataService);
```
donJoseLuis, line 85: what is the "check" signal?  Does the method throw if the check fails?

src/MonitoringFunctions/Models/HttpRequestLogEntry.cs
```diff
@@ -8,7 +9,7 @@ namespace MonitoringFunctions.Models
     /// <summary>
     /// Represents an http request event made from a function to be inserted into Kusto.
     /// </summary>
-    internal class HttpRequestLogEntry
+    internal class HttpRequestLogEntry : IKustoTableRow
```
donJoseLuis, line 12: could this be sealed?

src/MonitoringFunctions/Models/HttpRequestLogEntry.cs
```diff
@@ -19,7 +20,7 @@ internal class HttpRequestLogEntry
         [JsonProperty("requested_url"), JsonRequired]
         public string? RequestedUrl { get; set; }
 
-        [JsonProperty("http_status_code"), JsonRequired]
-        public int? HttpStatusCode { get; set; }
+        [JsonProperty("http_response_code"), JsonRequired]
+        public int? HttpResponseCode { get; set; }
```
donJoseLuis, line 24: consider overriding to string for models

src/MonitoringFunctions/Models/HttpRequestLogEntry.cs
donJoseLuis, line 12: if these will be short-lived instances, consider a struct (if then also implement IEquatable

src/MonitoringFunctions/Models/ScriptDryRunResult.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace MonitoringFunctions.Models
+{
+    internal class ScriptDryRunResult
+    {
+        public string? PrimaryUrl { get; set; }
+        public string? LegacyUrl { get; set; }
+    }
+}
```
donJoseLuis, line 10: consider sealing & overriding to string

src/MonitoringFunctions/Models/ScriptExecutionLogEntry.cs
```diff
@@ -0,0 +1,29 @@
...
+using MonitoringFunctions.DataService.Kusto;
+using Newtonsoft.Json;
+using System;
+
+namespace MonitoringFunctions.Models
+{
+    /// <summary>
+    /// Represents a script execution event to be inserted into Kusto.
+    /// </summary>
+    internal class ScriptExecutionLogEntry : IKustoTableRow
```
donJoseLuis, line 12: consider sealing and overriding to string

tests/MonitoringFunctions.Test/TestDirectJsonMappingResolver.cs
```diff
@@ -0,0 +1,102 @@
...
+    public class TestDirectJsonMappingResolver
+    {
+        private class BasicMappingModel : IKustoTableRow
+        {
+            public float someFloat { get; set; }
+
+            public string someString { get; set; }
+        }
+
+        private class CommonMappingModel : IKustoTableRow
```
donJoseLuis, line 26: this could be a child of ```ComplexMappingModel```

tests/MonitoringFunctions.Test/TestHelperMethods.cs
```diff
@@ -34,5 +36,43 @@ public async Task TestCheckUrlAccessFailure(string url)
                 // Test passed
             }
         }
+
+        [DataRow("-DryRun -c 3.0")]
+        [DataRow("-DryRun -c release/5.0.1xx-preview7")]
+        [DataRow("-DryRun -Version LTS")]
+        [TestMethod]
+        public async Task TestExecuteInstallScriptPs1(string cmdArgs = "-DryRun")
```
donJoseLuis, line 44: consider async suffix

src/MonitoringFunctions/DataService/DataServiceFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public IDataService GetDataService()
+        {
+            string? environment = Environment.GetEnvironmentVariable("AZURE_FUNCTIONS_ENVIRONMENT");
+
+            if(environment == "Development")
```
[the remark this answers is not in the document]
bozturkMSFT: There is no known value other than "Development". Enum wouldn't have anything but `Development`.

Also, the usage is quite simple here, in my opinion, to create an enum for it. We can consider this if we end up using it at more places.

src/MonitoringFunctions/DataService/Kusto/DirectJsonMappingResolver.cs
bozturkMSFT: There isn't a TypeArgumentException type in BCL, so I settled for ArgumentException.

src/MonitoringFunctions/DataService/Kusto/IKustoTableRow.cs
bozturkMSFT: I use this interface to refer to all the data classes we have. Otherwise, I would have to refer to them as `object` s. This makes the intention of the methods a bit more clear, in my opinion. It implies "this method requires a kusto table row, rather than any object" (even though any object should be good enough to be a kusto table row). 

Let's discuss this if you aren't sure that this is the right way of doing this, so that we can have the best design possible.

src/MonitoringFunctions/DataService/Kusto/KustoTable.cs
bozturkMSFT: Yes, this has been running on Azure for a couple days with no issues.
I think you can see that behavior if you forget to await the call, but otherwise, I would consider that a runtime bug. I would be very interested to see this, if you know of a sample that we can reproduce this with.

src/MonitoringFunctions/DataService/Providers/DummyDataService.cs
bozturkMSFT: What value would that add?

src/MonitoringFunctions/DataService/Providers/DummyDataService.cs
bozturkMSFT: Agreed

src/MonitoringFunctions/Functions/DryRunUrlChecker.cs
bozturkMSFT: Yes, exceptions report the failures. Updated the documentation to make it more clear.

tests/MonitoringFunctions.Test/TestHelperMethods.cs
bozturkMSFT: Isn't asking for Async suffix on test methods is the kind of bureaucracy that we are trying to avoid?

tests/MonitoringFunctions.Test/TestDirectJsonMappingResolver.cs
bozturkMSFT: Made `ComplexMappingModel` inherit from `CommonMappingModel` and modified the members a bit. This will better test the abilities of our column mapping resolver.

src/MonitoringFunctions/DataService/DataServiceFactory.cs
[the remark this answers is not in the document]
donJoseLuis: there is also production, right?  Consequentially, will we have regional PROD environments?  Such as US west, US east, ASIA, etc.?

src/MonitoringFunctions/DataService/DataServiceFactory.cs
donJoseLuis: i am a +1 on sealing.  Would like to understand what design do you have in mind to leaving it unsealed?

tests/MonitoringFunctions.Test/TestHelperMethods.cs
donJoseLuis: if there is a best practice, why make an exception for test code?  It's like you are purposely not trying to do your best in test code :)

donJoseLuis: this is minor, up to you to resolve as is (or do the right thing for the code, the team, the company and mankind)

src/MonitoringFunctions/Models/HttpRequestLogEntry.cs
donJoseLuis: please resolve at your prerogative

src/MonitoringFunctions/DataService/Kusto/DirectJsonMappingResolver.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    /// <summary>
+    /// Creates column mapping for objects whose json property names directly match the column name in the corresponding Kusto table.
+    /// </summary>
+    internal sealed class DirectJsonMappingResolver : IJsonColumnMappingResolver
+    {
+        public IEnumerable<ColumnMapping> GetColumnMappings<T>() where T : IKustoTableRow
+        {
+            DefaultContractResolver contractResolver = new DefaultContractResolver();
+            var contract = contractResolver.ResolveContract(typeof(T)) as JsonObjectContract;
```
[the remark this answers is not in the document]
donJoseLuis: resolution up to you on this one.

src/MonitoringFunctions/DataService/DataServiceFactory.cs
donJoseLuis: minor issue.  I'm championing for best practices and homogeneity in our team code base.

src/MonitoringFunctions/DataService/DataServiceFactory.cs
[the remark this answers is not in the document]
bozturkMSFT: No, nothing sets the environment variable to production. This whole thing is the courtesy of visual studio when running the functions locally. It's not used in Azure environment for instance (it returns empty string).

src/MonitoringFunctions/DataService/Kusto/DirectJsonMappingResolver.cs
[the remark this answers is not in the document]
bozturkMSFT: `typeof(T)` is used twice: one will execute always, the other one when throwing an exception. The latter is expected to run 1% of the time, maybe less. I think it may even hurt the performance to store the value of the type for a code path that will not execute for 99% of the time.

src/MonitoringFunctions/DataService/Providers/DummyDataService.cs
donJoseLuis: 1. public methods could check the state of the bool & not continue running if already disposed

https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose

bozturkMSFT: I agree that the proper way to implement dispose is to add a boolean to track the state. But, for this implementation, there are no resources to release. We are implementing `IDisposable` only to comply with the requirements of the interface.

I don't see how adding a boolean into the class and adding an if check that will probably be optimized away when compiled will benefit us. It will just make the code harder to read and maintain.

Leaving the method empty now doesn't mean a long term commitment. If we have some resources to free tomorrow, it will be pretty easy to add the boolean.

PR #67, opened 2020-08-07
ProxyBypassList must not be null

batzen:
ProxyBypassList must not be null.
Otherwise setting it on the proxy object throws an exception.


bozturkMSFT:
Hi @batzen, 

Apparently, we missed that setting `BypassList` to null is throwing an exception and it isn't mentioned in the official docs. 
I created an issue for a future fix in the libraries: https://github.com/dotnet/runtime/issues/40655.

 In the meantime, I will merge this PR to fix the issue.

PR #68, opened 2020-08-10
Get default proxy bypass list

amos402:
* https://github.com/dotnet/install-scripts/blob/ca03088d777fb308610b1c69999b9ecb9697d53a/src/dotnet-install.ps1#L262 needs a `ProxyBypassList`, but it didn't set after got the default proxy setting
* Relate with #67

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -245,6 +245,7 @@ function GetHTTPResponse([Uri] $Uri)
                     if($DefaultProxy -and (-not $DefaultProxy.IsBypassed($Uri))) {
                         $ProxyAddress = $DefaultProxy.GetProxy($Uri).OriginalString
                         $ProxyUseDefaultCredentials = $true
+                        $ProxyBypassList = ([System.Net.WebProxy]$DefaultProxy).BypassList
                     }
                 } catch {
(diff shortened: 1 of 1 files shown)


bozturkMSFT:
Hi @amos402,

 There is already a check for bypassed `Uri`s 2 lines above your edit:
https://github.com/dotnet/install-scripts/blob/ca03088d777fb308610b1c69999b9ecb9697d53a/src/dotnet-install.ps1#L245

 Does this not already cover your case? What issues are you having?

amos402:
It's not about the bypass checking, the reason it failed was caused it didn't set `ProxyBypassList` when it need a proxy, but it try to set the `BypassList` property after create the `WebProxy` object.
Here's a simple demonstration:
image

bozturkMSFT:
I see. Thank you for explaining.

We already plan to merge #67. That will also solve the problem you are having here. Can we close this PR or is there an edge case I am not seeing?

amos402:
Of course, that can solve most cases except the target URI have a redirection and the default proxy include that final destination, but that's out of the question.
Even the behavior may have a ambiguity(trying to get a default proxy but exclude its BypassList even the IsBypassed had been checked by that list). Well, since it's just internal tool, it's not a big deal. We can just close it.

PR #70, opened 2020-08-14
Monitoring functions for bash script

bozturkMSFT:
Delivers ADO work item [#1162628](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1162628)

Changes aim to add new functions to `-DryRun` bash script and publish the results to the database.
This includes:
- Added a new project `MonitoringFunctions.Linux` to maintain the functions that will run on Linux environment.
- Added a new project `MonitoringFunctions.Common` to share code between Windows and Linux functions.
- Moved some of the private methods out of `DryRunChecker` type into `HelperMethods` and `InstallScriptRunner` types under `Common` project.

Existing tests are shared between the two platforms. Which script to use for the tests depends on the OS you are running the test from.


grinrag (approved):

src/MonitoringFunctions.Linux/.gitignore
```diff
@@ -0,0 +1,264 @@
+## Ignore Visual Studio temporary files, build results, and
```
ostorc, line 1: Is there reason, why is this not part of main .gitignore?

ostorc (approved):

src/MonitoringFunctions.Linux/DryRunUrlChecker.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Extensions.Logging;
+
+namespace MonitoringFunctions.Linux
+{
+    /// <summary>
+    /// Runs the shell script in -DryRun mode and checks weather the generated links are accessible
```
ladipro, line 10: nit: Typo `weather`

src/MonitoringFunctions.Common/InstallScriptRunner.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <returns>Object containing primary and legacy runs</returns>
+        public static ScriptDryRunResult ParseDryRunOutput(string? output)
+        {
+            string primaryUrlIdentifier = "Primary named payload URL: ";
+            string legacyUrlIdentifier = "Legacy named payload URL: ";
+
+            ScriptDryRunResult result = new ScriptDryRunResult();
+
+            using StringStream stringStream = new StringStream(output);
+            using StreamReader streamReader = new StreamReader(stringStream);
```
ladipro, line 58: nit: I see it's pre-existing code but I'm curious why it's not using `StringReader` to parse the string.

```suggestion
            using StringReader streamReader = new StringReader(output);
```

ladipro (approved):
With the introduction of `MonitoringFunctions.Linux` shouldn't `MonitoringFunctions` be renamed to `MonitoringFunctions.Windows`? Thanks!

src/MonitoringFunctions.Linux/.gitignore
bozturkMSFT: I am not sure as these are auto-generated. But it maybe because different projects can have different ignore files and it is easier to maintain separate files instead of merging them all.

src/MonitoringFunctions.Common/InstallScriptRunner.cs
bozturkMSFT: `StringReader` only has a constructor that takes a string. That means we need to have the whole console output as a single string in memory. Not a huge issue, but I'm not sure if it is worth changing.

ladipro: We already have the entire output as a single string - passed in the `output` parameter. Hope I'm not reading it wrong.

bozturkMSFT: Ouch, you are right. I didn't look carefully, thought `output` was a stream. I merged this already but I will get that in the next PR.

PR #78, opened 2020-09-01
Function deployment to TEST environment and using TEST env DB

bozturkMSFT:
Delivers [#1157471](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1157471)

Build pipeline was modified to publish to the correct environment depending on the queue time variables.
Modified the functions to use the correct database for the selected environment.
Changed the permissions of bash script to allow running test on Linux.


src/MonitoringFunctions.Common/DataService/Providers/KustoDataService.cs
```diff
@@ -13,13 +13,19 @@ namespace MonitoringFunctions
...
+        private static readonly string? DatabaseName = Environment.GetEnvironmentVariable("kusto_db_name");
 
         private readonly KustoTable<HttpRequestLogEntry> _httpRequestLogsTable;
         private readonly KustoTable<ScriptExecutionLogEntry> _scriptExecutionLogsTable;
 
         internal KustoDataService()
         {
+            if (string.IsNullOrWhiteSpace(DatabaseName))
+            {
+                throw new ArgumentException($"{nameof(DatabaseName)} was not correctly configured. " +
```
ostorc, line 25: nit: Is `ArgumentException` appropriate here, since `DatabaseName` is not argument of constructor? Maybe use instead `InvalidOperationException`, citing documentation:
> The exception that is thrown when a method call is invalid for the object's current state.

ostorc (approved):

donJoseLuis (approved):
Would be great to get someone familiar with YAML take a loook

azure-pipelines.yml
```diff
@@ -113,42 +113,75 @@ stages:
...
+              ConnectionType: 'AzureRM'
+              # Name of the service connection that will give us access to Dotnet-Install-Monitoring resource group
+              # This is where the windows based Azure Functions are allocated.
+              azureSubscription: 'InstallScripts_KeyVault'
+              appType: 'functionApp'
+              WebAppName: dotnet-install-daemon-functions
+              deployToSlotOrASE: true
+              ResourceGroupName: 'Dotnet-Install-Monitoring'
+              SlotName: production
+              packageForLinux: '$(System.DefaultWorkingDirectory)/build_windows_$(Build.BuildId).zip'
```
ladipro, line 173: This line refers to both Linux and Windows. Is it expected?

ladipro (approved):
Looks good to my yml-untrained eye!

grinrag (approved):

azure-pipelines.yml
bozturkMSFT: `packageForLinux` is just an alias for `Package`, but I agree that it looks wrong.
I replaced it with `Package` now.

PR #82, opened 2020-09-16
Monitoring functions do not properly trigger

### LINKED ISSUE #81: Monitoring functions do not properly trigger
Our functions on azure are not triggering properly. The symptoms are as follows

Function App Name | Issue
-- | --
dotnet-install-daemon-functions | Irregular triggering. Some of the triggerings are being skipped (Instead of every 30 minutes, functions are triggering every 1 or 1.5 hours)
dotnet-install-daemon-functions/test | Irregular triggering. Prints an error at each run: An attempt was made to access a socket in a way forbidden by its access permissions.
dotnet-install-daemon-functions-linux | Publishing succeeds, but functions are not available in the app.
dotnet-install-daemon-functions-linux-test | Not triggering at all.

Fix the issues so that the functions are triggering as expected.

bozturkMSFT:
Fixes #81 

Function apps were removed and recreated in Azure.
Publishing profiles were updated to publish to the new resources.
App names are updated. Therefore, build pipeline is also updated with the new names.
Unrelated: Windows functions are placed into correct namespace.


AR-May (approved):
LGTM

vlada-shubina (approved):

Issue #38, opened 2020-05-28
Update repo in script's exception URLs


donJoseLuis:
Change repo from sdk to install-scripts under the dotnet owner in scripts' exception messages.

Example
throw "Architecture not supported. If you think this is a bug, report it at https://github.com/dotnet/sdk/issues" should be changed to throw "Architecture not supported. If you think this is a bug, report it at https://github.com/dotnet/install-scripts/issues"

AR-May:
Fixed

PR #86, opened 2020-10-06
Change repo from sdk to install-scripts in scripts

AR-May:
Delivers #38.  Change repo from sdk to install-scripts in scripts' exception messages.

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -195,7 +195,7 @@ function Get-CLIArchitecture-From-Architecture([string]$Architecture) {
         { $_ -eq "x86" } { return "x86" }
         { $_ -eq "arm" } { return "arm" }
         { $_ -eq "arm64" } { return "arm64" }
-        default { throw "Architecture not supported. If you think this is a bug, report it at https://github.com/dotnet/sdk/issues" }
+        default { throw "Architecture '$Architecture' not supported. If you think this is a bug, report it at https://github.com/dotnet/install-scripts/issues" }
     }
 }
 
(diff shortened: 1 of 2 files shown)


AR-May:
I have found only one message with a link to https://github.com/dotnet/sdk/issues which should be replaced.

bozturkMSFT (approved):

src/dotnet-install.ps1
```diff
@@ -195,7 +195,7 @@ function Get-CLIArchitecture-From-Architecture([string]$Architecture) {
         { $_ -eq "x86" } { return "x86" }
         { $_ -eq "arm" } { return "arm" }
         { $_ -eq "arm64" } { return "arm64" }
-        default { throw "Architecture not supported. If you think this is a bug, report it at https://github.com/dotnet/sdk/issues" }
+        default { throw "Architecture not supported. If you think this is a bug, report it at https://github.com/dotnet/install-scripts/issues" }
```
donJoseLuis, line 198: would it add value to include the specified architecture which is not supported?

@AR-May  I know, not your change...while we're here, would the above make our exception better?

@bozturkMSFT , what do you think?

donJoseLuis (approved):
minor suggestion

src/dotnet-install.ps1
AR-May: For me, it sounds reasonable. I can edit the message like that: "Architecture '$Architecture' not supported.".
@bozturkMSFT, what do you think of this?

grinrag (approved):

src/dotnet-install.ps1
bozturkMSFT: Not a "must", in my opinion, but it would be nice to have.

PR #90, opened 2020-10-23
Fix SDK version parsing from global.json in install script on macOS

kondratyev-nv:
For #7 

Running install script `scripts/obtain/dotnet-install.sh` with `jsonfile` argument works on macOS only when `sdk` section has a single property `version`. When `sdk` section has multiple properties, for example,
```json
"sdk": {
    "version":  "5.0.100-preview.6.20310.4",
    "allowPrerelease": true,
    "rollForward": "major"
  }
```
a version is detected as `5.0.100-preview.6.20310.4allowPrerelease:truerollForward:major`.
Also, in a case when `version` is not the first property in `sdk` a version can not be detected returning an error ``dotnet_install: Error: Unable to find the SDK:version node in `global.json` ``.

The problem seems to be in line:
```
sdk_list="$(echo -e "${sdk_list}" | tr -d '[[:space:]]')"
```
On macOS, echo does not have a `-e` argument, and all lines of key-value pairs in `sdk_list` are merged into a single line. So, removing it seems to fix it. I don't have a Linux to test the changes, but I hope CI will catch any issues on Linux.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -468,7 +468,6 @@ parse_jsonfile_for_version() {
     sdk_list=$(echo $sdk_section | awk -F"[{}]" '{print $2}')
     sdk_list=${sdk_list//[\" ]/}
     sdk_list=${sdk_list//,/$'\n'}
-    sdk_list="$(echo -e "${sdk_list}" | tr -d '[[:space:]]')"
 
     local version_info=""
     while read -r line; do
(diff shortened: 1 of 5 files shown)


AR-May (approved):
LGTM

AR-May:
I tested changes on both Mac and Linux, the fix works fine for me. @bozturkMSFT, please, review too.

bozturkMSFT (approved):
LGTM

Issue #72, opened 2020-08-20
Installation on Alpine misses some prerequisites, so dotnet won't run


danmosemsft:
Alpine 3.11.3 on WSL2
```
danmose3:~/git/tmp# cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.11.3
PRETTY_NAME="Alpine Linux v3.11"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"
```
Installing 5.0 with dotnet-install.sh leaves me with a broken `dotnet`:
```
[2020/08/20 21:03:10][INFO] $ /root/git/performance/tools/dotnet/x64/dotnet-install.sh -InstallDir /root/git/performance/tools/dotnet/x64 -Architecture x64 -Channel master
[2020/08/20 21:03:11][INFO] dotnet_install: Warning: Unable to locate libintl. Probable prerequisite missing; install libintl (or gettext).
[2020/08/20 21:03:11][INFO] dotnet_install: Warning: Unable to locate zlib. Probable prerequisite missing; install zlib.
[2020/08/20 21:03:11][INFO] dotnet_install: Warning: Unable to locate libicu. Probable prerequisite missing; install libicu.
[2020/08/20 21:03:11][INFO] dotnet_install: Warning: Unable to locate liblttng. Probable prerequisite missing; install libcurl.
[2020/08/20 21:03:11][INFO] dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.100-rc.1.20420.14/dotnet-sdk-5.0.100-rc.1.20420.14-linux-musl-x64.tar.gz
...
[2020/08/20 21:03:14][INFO]
[2020/08/20 21:03:15][INFO] dotnet-install: Extracting zip from https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.100-rc.1.20420.14/dotnet-sdk-5.0.100-rc.1.20420.14-linux-musl-x64.tar.gz
[2020/08/20 21:03:18][INFO] dotnet-install: Adding to current process PATH: `/root/git/performance/tools/dotnet/x64`. Note: This change will be visible only when sourcing script.
[2020/08/20 21:03:18][INFO] dotnet-install: Installation finished successfully.
...
[2020/08/20 21:03:26][INFO] $ dotnet --info
[...]

Issue #36, opened 2020-05-28
Requiring tools (libcurl) that are not needed


donJoseLuis:
Separated from couple of issues bundled in #6 

exit criteria
1. scripts ONLY perform checks for required dependencies

PR #91, opened 2020-10-27
Removing dependencies check, adding a link to documentation page

AR-May:
Concerns issues #6, #36, #72.

The dependencies check is currently broken: 
- it is out of date, 
- it is not full (and can't be full): for different versions, for runtime/SDK, there might be different sets of dependencies, 
- the warnings are misleading, as they do not specify that it is a library name, not a package name, which has not been found.

To support these checks properly (more/less full and not out of date) we would need too big of an effort. It is doubtful that some checks always are better than none: we are worried that the customer would see the warning, add the library, and would not understand why dotnet still does not work, as a result of incompleteness of a check.

Based on these, we decided to remove them and instead always ask the customer to visit the documentation page and to check the dependencies listed there.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -241,42 +241,6 @@ check_min_reqs() {
     return 0
 }
 
-check_pre_reqs() {
-    eval $invocation
-
-    if [ "${DOTNET_INSTALL_SKIP_PREREQS:-}" = "1" ]; then
-        return 0
-    fi
-
-    if [ "$(uname)" = "Linux" ]; then
-        if is_musl_based_distro; then
-            if ! command -v scanelf > /dev/null; then
-                say_warning "scanelf not found, please install pax-utils package."
-                return 0
-            fi
-            LDCONFIG_COMMAND="scanelf --ldpath -BF '%f'"
-            [ -z "$($LDCONFIG_COMMAND 2>/dev/null | grep libintl)" ] && say_warning "Unable to locate libintl. Probable prerequisite missing; install libintl (or gettext)."
-        else
-            if [ ! -x "$(command -v ldconfig)" ]; then
-                say_verbose "ldconfig is not in PATH, trying /sbin/ldconfig."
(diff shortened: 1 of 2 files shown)


donJoseLuis (approved):
Thanks for moving this forward. We definitely need @bozturkMSFT 's signoff.

src/dotnet-install.sh
```diff
@@ -1130,4 +1093,5 @@ else
     say "Binaries of dotnet can be found in $bin_path"
 fi
 
+say "Please check the full list of dependencies in the documentation: https://docs.microsoft.com/en-us/dotnet/core/install"
```
bozturkMSFT, line 1096: Would people clicking the link get lost in the pages before they can reach de "dependencies" section?

Maybe we can prevent that by being a bit more explicit, like:
"To check the full list of dependencies, please go to the following link, select your operating system and check the \"Depedencies\" section: https://docs.microsoft.com/en-us/dotnet/core/install"

I am not sure this is the proper way to escape quotes.

grinrag (approved):
LGTM

src/dotnet-install.sh
AR-May: I agree we may be more verbose. However, I would like to make a message that looks like a necessary action item. To be seen it also should be either quite short or colored. The script is for developer's use, I hope that they can work with documentation and find information there. I can suggest modifying the message like that:
```
Please check the full list of dependencies for your operating system in the documentation: https://docs.microsoft.com/en-us/dotnet/core/install
```

AR-May: However, I thought that maybe we should explain that the script does not resolve dependencies. So, we anyway need to be verbose. Here is the next proposal, two lines:
```
Please note the script does not resolve dependencies during installation.
To check the list of dependencies, go to https://docs.microsoft.com/en-us/dotnet/core/install, select your operating system, and check the "Dependencies" section.
```

bozturkMSFT (approved):

Issue #43, opened 2020-06-11
Review documentation for when the install scripts are to be used


vitek-karas:
See https://github.com/dotnet/docs/issues/18715 for the detailed discussion of the problem the current behavior causes to users.

donJoseLuis:
@vitek-karas thanks for your issue.  We can get to this next week due to our commitments for the current week.

AR-May:
Added warnings concerning the purpose of the scripts.

PR #94, opened 2020-11-03
Add info messages concerning install scripts purpose

AR-May:
Delivers issue #43. Adding a warning so that people who want to install a runtime to run apps should use other installation options.

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -635,6 +635,11 @@ function Prepend-Sdk-InstallRoot-To-Path([string]$InstallRoot, [string]$BinFolde
     }
 }
 
+Say "Please note that the intended use of this script is for Continuous Integration (CI) scenarios, where:"
+Say "- The SDK needs to be installed without user interaction and without admin rights."
+Say "- The SDK installation doesn't need to persist across multiple CI runs."
+Say "To set up a development environment or to run apps, use installers rather than this script. Visit https://dotnet.microsoft.com/download to get the installer.`r`n"
+
 $CLIArchitecture = Get-CLIArchitecture-From-Architecture $Architecture
 $SpecificVersion = Get-Specific-Version-From-Version -AzureFeed $AzureFeed -Channel $Channel -Version $Version -JSonFile $JSonFile
 $DownloadLink, $EffectiveVersion = Get-Download-Link -AzureFeed $AzureFeed -SpecificVersion $SpecificVersion -CLIArchitecture $CLIArchitecture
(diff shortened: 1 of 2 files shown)


src/dotnet-install.ps1
```diff
@@ -145,6 +145,18 @@ function Say-Verbose($str) {
     }
 }
 
+function Say-Text-Separator() {
```
donJoseLuis, line 148: I'm unsure on the use case requiring this uncommon function.  The syntax appears correct, though it's difficult to sign off on how it is needed.

src/dotnet-install.ps1
```diff
@@ -635,6 +647,12 @@ function Prepend-Sdk-InstallRoot-To-Path([string]$InstallRoot, [string]$BinFolde
     }
 }
 
+Say "Please note that the intended use of this script is for Continuous Integration (CI) scenarios, where:"
+Say "- The SDK needs to be installed without user interaction and without admin rights."
+Say "- The SDK installation doesn't need to persist across multiple CI runs."
+Say "To set up a development environment or to run apps, use installers rather than this script. Visit https://dotnet.microsoft.com/download to get the installer."
+Say-Text-Separator
```
donJoseLuis, line 654: minor: Does System.Environment.NewLine not get the job done instead of needing `Say-Text-Separator`?

donJoseLuis (approved):
some minor comments

src/dotnet-install.ps1
bozturkMSFT: I understand that the function is used to print a new line without prefixing it with `dotnet-install:`. However, we can achieve the same effect if we appended a new line to the previous message, right? The last line would look like this:

```
Say "To set up a dev... ... to get the installer.`r`n"
```

Can we get rid of `Say-Text-Separator` afterwards?

src/dotnet-install.ps1
AR-May: There is another function, `Say`, which writes whatever you wish there with the prefix "dotnet-install". I do not wish to have this prefix with a blank line, so I decided not to reuse it. I do not want to modify the function `Say` to be able to write a blank line (or in other words, a line with any prefix), as it will involve too wide changes in the code, I will need to pass "dotnet-install" everywhere else.  At the same time I do wish this separator to be written in the same stream as the whole message. So, I copied the logic of the function `Say` to get this.

src/dotnet-install.ps1
AR-May: That is a good idea, yes, this will give the needed effect. I'll modify scripts as @bozturkMSFT suggests.

bozturkMSFT (approved):

PR #108, opened 2020-11-12
Send incident alerts to Teams

bekir-ozturk:
Changes in this PR posts a card to a teams channel whenever a new work item is created for an incident.

### DIFF
--- a/src/MonitoringFunctions.Common/TeamsConnector.cs
+++ b/src/MonitoringFunctions.Common/TeamsConnector.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+using System;
 using System.Net.Http;
 using System.Text;
 using System.Threading;
@@ -6,13 +8,15 @@ using System.Threading.Tasks;
 
 namespace MonitoringFunctions.Common
 {
+    /// <summary>
+    /// Represents a connector for MS Teams.
+    /// </summary>
     internal sealed class TeamsConnector
     {
-        private static readonly HttpClient _httpClient = new HttpClient();
-
+        private static readonly HttpClient HttpClient = new HttpClient();
+        private const string IncidentCardFormat = "{{\"@context\":\"https://schema.org/extensions\",\"@type\":\"MessageCard\",\"themeColor\":\"FF5555\",\"title\":\"{0}\",\"text\":\"{1}\",\"potentialAction\":[{{\"@type\":\"OpenUri\",\"name\":\"Go To Work Item\",\"targets\":[{{\"os\":\"default\",\"uri\":\"{2}\"}}]}}]}}";
+        
         private string _webhookUrl;
-
-        private const string _incidentCardFormat = "{{\"@context\":\"https://schema.org/extensions\",\"@type\":\"MessageCard\",\"themeColor\":\"FF5555\",\"title\":\"{0}\",\"text\":\"{1}\",\"potentialAction\":[{{\"@type\":\"OpenUri\",\"name\":\"Go To Work Item\",\"targets\":[{{\"os\":\"default\",\"uri\":\"{2}\"}}]}}]}}";
 
         /// <summary>
         /// Creates a new instance of type <see cref="TeamsConnector"/>.
@@ -41,10 +45,10 @@ namespace MonitoringFunctions.Common
         /// <remarks>It is safe to call this method on the same object simultaneously from multiple threads.</remarks>
         public async Task SendIncidentCard(string title, string description, string workItemUrl, CancellationToken cancellationToken = default)
         {
-            string payload = string.Format(_incidentCardFormat, title, description, workItemUrl);
+            string payload = string.Format(IncidentCardFormat, title, description, workItemUrl);
             StringContent payloadContent = new StringContent(payload, Encoding.UTF8);
 
-            HttpResponseMessage response = await _httpClient.PostAsync(_webhookUrl, payloadContent, cancellationToken);
+            HttpResponseMessage response = await HttpClient.PostAsync(_webhookUrl, payloadContent, cancellationToken);
 
             if(!response.IsSuccessStatusCode)
             {
--- a/src/MonitoringFunctions.Common/TeamsConnector.cs
+++ b/src/MonitoringFunctions.Common/TeamsConnector.cs
@@ -43,7 +43,7 @@ namespace MonitoringFunctions.Common
         /// <returns>Task that tracks the completion of the operation.
         /// Errors are communicated through exceptions.</returns>
         /// <remarks>It is safe to call this method on the same object simultaneously from multiple threads.</remarks>
-        public async Task SendIncidentCard(string title, string description, string workItemUrl, CancellationToken cancellationToken = default)
+        public async Task SendIncidentCardAsync(string title, string description, string workItemUrl, CancellationToken cancellationToken = default)
         {
             string payload = string.Format(IncidentCardFormat, title, description, workItemUrl);
             using StringContent payloadContent = new StringContent(payload, Encoding.UTF8);
@@ -52,7 +52,9 @@ namespace MonitoringFunctions.Common
 
             if(!response.IsSuccessStatusCode)
             {
-                throw new Exception($"Sending incident card has failed with http status code {response.StatusCode}.");
+                string responseContent = await response.Content.ReadAsStringAsync();
+                throw new HttpRequestException($"Sending incident card has failed with http status code {response.StatusCode}."
+                    + " Response content: " + responseContent);
             }
         }
     }
--- a/src/MonitoringFunctions.Common/TeamsConnector.cs
+++ b/src/MonitoringFunctions.Common/TeamsConnector.cs
@@ -46,9 +46,9 @@ namespace MonitoringFunctions.Common
         public async Task SendIncidentCard(string title, string description, string workItemUrl, CancellationToken cancellationToken = default)
         {
             string payload = string.Format(IncidentCardFormat, title, description, workItemUrl);
-            StringContent payloadContent = new StringContent(payload, Encoding.UTF8);
+            using StringContent payloadContent = new StringContent(payload, Encoding.UTF8);
 
-            HttpResponseMessage response = await HttpClient.PostAsync(_webhookUrl, payloadContent, cancellationToken);
+            using HttpResponseMessage response = await HttpClient.PostAsync(_webhookUrl, payloadContent, cancellationToken);
 
             if(!response.IsSuccessStatusCode)
             {
--- a/src/MonitoringFunctions.Common/TeamsConnector.cs
+++ b/src/MonitoringFunctions.Common/TeamsConnector.cs
@@ -1,0 +1,55 @@
+﻿using System;
+using System.Net.Http;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace MonitoringFunctions.Common
+{
+    internal sealed class TeamsConnector
+    {
+        private static readonly HttpClient _httpClient = new HttpClient();
+
+        private string _webhookUrl;
+
+        private const string _incidentCardFormat = "{{\"@context\":\"https://schema.org/extensions\",\"@type\":\"MessageCard\",\"themeColor\":\"FF5555\",\"title\":\"{0}\",\"text\":\"{1}\",\"potentialAction\":[{{\"@type\":\"OpenUri\",\"name\":\"Go To Work Item\",\"targets\":[{{\"os\":\"default\",\"uri\":\"{2}\"}}]}}]}}";
+
+        /// <summary>
+        /// Creates a new instance of type <see cref="TeamsConnector"/>.
+        /// </summary>
+        /// <param name="webhookUrl">Webhook URL that can be used to access this connector on Teams.</param>
+        /// <exception cref="ArgumentException">If the supplied webhook url is null or whitespace.</exception>
+        public TeamsConnector(string webhookUrl)
+        {
+            if(string.IsNullOrWhiteSpace(webhookUrl))
+            {
+                throw new ArgumentException($"{webhookUrl} can not be null or empty.");
+            }
+
+            _webhookUrl = webhookUrl;
+        }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/MonitoringFunctions.Common/TeamsConnector.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Threading.Tasks;
+
+namespace MonitoringFunctions.Common
+{
+    /// <summary>
+    /// Represents a connector for MS Teams.
+    /// </summary>
+    internal sealed class TeamsConnector
+    {
+        private static readonly HttpClient HttpClient = new HttpClient();
```
donJoseLuis, line 16: `HttpClient` is disposable.

https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.dispose?view=net-5.0

Making matters more interesting, [C#: HttpClient should NOT be disposed](https://medium.com/@nuno.caneco/c-httpclient-should-not-be-disposed-or-should-it-45d2a8f568bc)

src/MonitoringFunctions.Common/TeamsConnector.cs
```diff
@@ -0,0 +1,59 @@
...
+        private string _webhookUrl;
+
+        /// <summary>
+        /// Creates a new instance of type <see cref="TeamsConnector"/>.
+        /// </summary>
+        /// <param name="webhookUrl">Webhook URL that can be used to access this connector on Teams.</param>
+        /// <exception cref="ArgumentException">If the supplied webhook url is null or whitespace.</exception>
+        public TeamsConnector(string webhookUrl)
+        {
+            if(string.IsNullOrWhiteSpace(webhookUrl))
```
donJoseLuis, line 28: consider:
`_webhookUrl = (string.IsNullOrWhiteSpace(webhookUrl) ?  throw new ArgumentNullException(paramName: nameof(webhookUrl)) : webhookUrl;`

src/MonitoringFunctions.Common/TeamsConnector.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <param name="cancellationToken">Token to cancel the operation.</param>
+        /// <returns>Task that tracks the completion of the operation.
+        /// Errors are communicated through exceptions.</returns>
+        /// <remarks>It is safe to call this method on the same object simultaneously from multiple threads.</remarks>
+        public async Task SendIncidentCard(string title, string description, string workItemUrl, CancellationToken cancellationToken = default)
+        {
+            string payload = string.Format(IncidentCardFormat, title, description, workItemUrl);
+            StringContent payloadContent = new StringContent(payload, Encoding.UTF8);
+
+            HttpResponseMessage response = await HttpClient.PostAsync(_webhookUrl, payloadContent, cancellationToken);
```
donJoseLuis, line 51: The HttpResponseMessage Class is disposable

https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpresponsemessage?view=net-5.0

src/MonitoringFunctions.Common/TeamsConnector.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <param name="description">Description text on the card.</param>
+        /// <param name="workItemUrl">Link to the work item of the incident.</param>
+        /// <param name="cancellationToken">Token to cancel the operation.</param>
+        /// <returns>Task that tracks the completion of the operation.
+        /// Errors are communicated through exceptions.</returns>
+        /// <remarks>It is safe to call this method on the same object simultaneously from multiple threads.</remarks>
+        public async Task SendIncidentCard(string title, string description, string workItemUrl, CancellationToken cancellationToken = default)
+        {
+            string payload = string.Format(IncidentCardFormat, title, description, workItemUrl);
+            StringContent payloadContent = new StringContent(payload, Encoding.UTF8);
```
donJoseLuis, line 49: The StringContent class is disposable

https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpcontent.dispose?view=net-5.0#System_Net_Http_HttpContent_Dispose

src/MonitoringFunctions.Windows/Functions/WorkItemCreator.cs
```diff
@@ -64,6 +66,27 @@ internal static class WorkItemCreator
...
+            TeamsConnector? teamsConnector = null;
+
+            if (data.Tags != null
+                && data.Tags.ContainsKey("notifyTeams")
+                && bool.TryParse(data.Tags["notifyTeams"], out bool notifyTeams)
+                && notifyTeams)
+            {
+                if (string.IsNullOrWhiteSpace(TeamsWebhookUrl))
+                {
+                    log.LogError("Teams notifications were requested, but no webhook url was provided."
```
donJoseLuis, line 78: super minor: a plus if the 2 strings in the log called could be turned into 1

src/MonitoringFunctions.Windows/Functions/WorkItemCreator.cs
```diff
@@ -64,6 +66,27 @@ internal static class WorkItemCreator
             string alertMessage = data.Message ?? "Alert triggered";
             int alertingMonitorCount = data.MatchingAlerts?.Length ?? 0;
             List<WorkItem> createdWorkItems = new List<WorkItem>();
+            TeamsConnector? teamsConnector = null;
```
donJoseLuis, line 69: heads up, if you do decide to make TeamConnector implement IDisposable, then you'd want to properly dispose this instance.

donJoseLuis (changes requested):
Being part of a service, we should aggressively prevent potential leaks.

src/MonitoringFunctions.Common/TeamsConnector.cs
nohwnd: Argument null imho means it is null, not empty or whitespace. The current exception is correct.

bozturkMSFT: About the styling suggestion: Though I like the one-line alternative in some cases, I think it is usually making the code harder to read. [Other examples](https://github.com/dotnet/runtime/search?p=9&q=argumentnullexception&type=code) I could find in the libraries are also styled this way. So I will keep this as is.

src/MonitoringFunctions.Common/TeamsConnector.cs
bozturkMSFT: I decided to keep it undisposed as discussed in the article.

src/MonitoringFunctions.Windows/Functions/WorkItemCreator.cs
bozturkMSFT: According to [Constant Expressions](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#constant-expressions) section of C# language specification, strings here will be concatenated at compile time (related SO post [here](https://stackoverflow.com/a/288802)). I will keep it as is to help readability.

src/MonitoringFunctions.Windows/Functions/WorkItemCreator.cs
bozturkMSFT: Thanks. `TeamsConnector` doesn't have any disposable instance members at the moment. So this shouldn't be fine.

src/MonitoringFunctions.Common/TeamsConnector.cs
donJoseLuis, line 28: very minor: consider a consolidated equivalent statement

```_webhookUrl = string.IsNullOrWhiteSpace(webhookUrl) ? throw new ArgumentException($"{webhookUrl} can not be null or empty.") : webhookUrl;```

File: src/MonitoringFunctions.Common/TeamsConnector.cs [hunk 7]
```diff
@@ -0,0 +1,59 @@
...
+        public async Task SendIncidentCard(string title, string description, string workItemUrl, CancellationToken cancellationToken = default)
+        {
+            string payload = string.Format(IncidentCardFormat, title, description, workItemUrl);
+            using StringContent payloadContent = new StringContent(payload, Encoding.UTF8);
+
+            using HttpResponseMessage response = await HttpClient.PostAsync(_webhookUrl, payloadContent, cancellationToken);
+
+            if(!response.IsSuccessStatusCode)
+            {
+                throw new Exception($"Sending incident card has failed with http status code {response.StatusCode}.");
```
donJoseLuis, line 55: minor: would including the response payload, in addition to the http status code, add value to the exception catcher?

donJoseLuis, line 55: very minor:  is type ```Exception``` the best Exception for an unsuccessful status?

src/MonitoringFunctions.Common/TeamsConnector.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace MonitoringFunctions.Common
+{
+    /// <summary>
+    /// Represents a connector for MS Teams.
+    /// </summary>
+    internal sealed class TeamsConnector
+    {
+        private static readonly HttpClient HttpClient = new HttpClient();
+        private const string IncidentCardFormat = "{{\"@context\":\"https://schema.org/extensions\",\"@type\":\"MessageCard\",\"themeColor\":\"FF5555\",\"title\":\"{0}\",\"text\":\"{1}\",\"potentialAction\":[{{\"@type\":\"OpenUri\",\"name\":\"Go To Work Item\",\"targets\":[{{\"os\":\"default\",\"uri\":\"{2}\"}}]}}]}}";
```
donJoseLuis, line 17: minor: would it avoid tech dept to instead of creating payload via parameter formatting to instead 
1. create a type, 
1. instantiate it & initialize its' state
1. then serialize it?

src/MonitoringFunctions.Windows/Functions/WorkItemCreator.cs
```diff
@@ -101,6 +124,18 @@ internal static class WorkItemCreator
 
                 log.LogInformation(successMessage);
                 createdWorkItems.Add(workItem);
+
+                if (teamsConnector != null)
+                {
+                    try
+                    {
+                        await teamsConnector.SendIncidentCard("There is a new issue with the install scripts.", title, workItemUrl);
```
donJoseLuis, line 132: consider configuring the awaitable to ```false```

donJoseLuis, line 132: minor: consider creating a cancellation token encapsulating a targeted time out & pass it to the call

donJoseLuis (changes requested):
thanks for the changes, a few comments

src/MonitoringFunctions.Common/TeamsConnector.cs
donJoseLuis: Also, consider an "Async" suffix for naming async methods

src/MonitoringFunctions.Common/TeamsConnector.cs
nohwnd: I think you could also move this definition directly into the SendIncidentCard method and use an anonymous type.

bozturkMSFT: Considering my availability this week, I would like to keep this as is at this time.
If we need to make a change next time, we can update as suggested.

src/MonitoringFunctions.Windows/Functions/WorkItemCreator.cs
```diff
@@ -38,7 +38,11 @@ internal static class WorkItemCreator
             [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
             ILogger log)
         {
-            using StreamReader requestStream = new StreamReader(req.Body);
+            // Function run shouldn't take longer than 3 minutes.
+            TimeSpan maxExecutionDuration = new TimeSpan(0, minutes: 3, 0);
```
nohwnd, line 42: nit:

```suggestion
            TimeSpan maxExecutionDuration =  TimeSpan.FromMinutes(3);
```

src/MonitoringFunctions.Common/TeamsConnector.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            string payload = string.Format(IncidentCardFormat, title, description, workItemUrl);
+            using StringContent payloadContent = new StringContent(payload, Encoding.UTF8);
+
+            using HttpResponseMessage response = await HttpClient.PostAsync(_webhookUrl, payloadContent, cancellationToken);
+
+            if(!response.IsSuccessStatusCode)
+            {
+                string responseContent = await response.Content.ReadAsStringAsync();
+                throw new HttpRequestException($"Sending incident card has failed with http status code {response.StatusCode}."
```
donJoseLuis, line 56: minor: consider setting the HttpRequestException.StatusCode Property

https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httprequestexception.statuscode?view=net-5.0

src/MonitoringFunctions.Common/TeamsConnector.cs
donJoseLuis: minor: the "I'll do it next time" bit usually = **tech-debt**.  Unless there is an urgency to finish, please consider following @nohwnd 's advise.

src/MonitoringFunctions.Windows/Functions/WorkItemCreator.cs
```diff
@@ -29,13 +30,19 @@ internal static class WorkItemCreator
...
         private static readonly IncidentSerializer IncidentSerializer = new IncidentSerializer();
 
         [FunctionName("WorkItemCreator")]
         public static async Task<IActionResult> Run(
             [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
             ILogger log)
         {
+            // Function run shouldn't take longer than 3 minutes.
+            TimeSpan maxExecutionDuration = TimeSpan.FromMinutes(3);
+            CancellationTokenSource cancellationTokenSource = new CancellationTokenSource(maxExecutionDuration);
```
donJoseLuis, line 43: token sources are disposable

donJoseLuis (changes requested):
Looks great, happy to see our integration onto Teams.

bekir-ozturk:
This wont be merged as there will be a new implementation in v2 monitoring. Closing.

Issue #136, opened 2021-02-05
Unbound variable error from the latest version of dotnet-install.sh


ujpandey:
The script exits with error due to an unbound variable. The code there seems to be assuming that the variable is set but that's only true if the download failed [as of this commit](https://github.com/dotnet/install-scripts/commit/c9e5e27e1ccfc827daf613c3eebbd860e8ea7e47) which unsets the variables explicitly when starting the download.

On MacOS:
```
./dotnet-install.sh -c Current
dotnet-install: Note that the intended use of this script is for Continuous Integration (CI) scenarios, where:
dotnet-install: - The SDK needs to be installed without user interaction and without admin rights.
dotnet-install: - The SDK installation doesn't need to persist across multiple CI runs.
dotnet-install: To set up a development environment or to run apps, use installers rather than this script. Visit https://dotnet.microsoft.com/download to get the installer.

dotnet-install: Downloading primary link https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.102/dotnet-sdk-5.0.102-osx-x64.tar.gz
./dotnet-install.sh: line 908: http_code: unbound variable
```

On Linux:
```
./dotnet-install.sh -c Current
dotnet-install: Note that the intended use of this script is for Continuous Integration (CI) scenarios, where:
dotnet-install: - The SDK needs to be installed without user interaction and without admin rights.
dotnet-install: - The SDK installation doesn't need to persist across multiple CI runs.
dotnet-install: To set up a development environment or to run apps, use installers rather than this script. Visit https://dotnet.microsoft.com/download to get the installer.

dotnet-install: Downloading primary link https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.102/dotnet-sdk-5.0.102-linux-x64.tar.gz
./dotnet-install.sh: line 908: http_code: unbound variable
```

BruceForstall:
@bozturkMSFT

danmosemsft:
@bozturkMSFT it would be good to understand how this bug made it through validation, as it is taking out our builds, what are your thoughts?  Perhaps there's some more validation we can add.

dougbu:
Offline is fine but I would appreciate a root cause analysis, not of the bug itself but of the rollout containing the bug.

dave-yotta:
This has broken our deployments and halted building images in dev - when will the fix be rolled out?

dave-yotta:
[...]

PR #137, opened 2021-02-05
Consult http_code and download_error_msg only for failed downloads.

ujpandey:
Should fix https://github.com/dotnet/install-scripts/issues/136

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -489,7 +489,7 @@ get_specific_version_from_version() {
     local json_file="$5"
 
     if [ -z "$json_file" ]; then
-        if [[ "$version" == "latest" ]]; then 
+        if [[ "$version" == "latest" ]]; then
             local version_info
             version_info="$(get_latest_version_info "$azure_feed" "$channel" "$normalized_architecture" false)" || return 1
             say_verbose "get_specific_version_from_version: version_info=$version_info"
@@ -522,7 +522,7 @@ construct_download_link() {
     local specific_version="${4//[$'\t\r\n']}"
     local specific_product_version="$(get_specific_product_version "$1" "$4")"
     local osname="$5"
-    
+
     local download_link=null
     if [[ "$runtime" == "dotnet" ]]; then
         download_link="$azure_feed/Runtime/$specific_version/dotnet-runtime-$specific_product_version-$osname-$normalized_architecture.tar.gz"
@@ -542,7 +542,7 @@ construct_download_link() {
 # azure_feed - $1
 # specific_version - $2
 get_specific_product_version() {
-    # If we find a 'productVersion.txt' at the root of any folder, we'll use its contents 
+    # If we find a 'productVersion.txt' at the root of any folder, we'll use its contents
     # to resolve the version of what's in the folder, superseding the specified version.
     eval $invocation
 
@@ -753,9 +753,8 @@ download() {
         elif machine_has "wget"; then
             downloadwget "$remote_path" "$out_path" || failed=true
         else
-            unset http_code
-            download_error_msg="Missing dependency: neither curl nor wget was found."
-            break
+            say_err "Missing dependency: neither curl nor wget was found."
+            exit 1
         fi
 
         if [ "$failed" = false ] || [ $attempts -ge 3 ] || { [ ! -z $http_code ] && [ $http_code = "404" ]; }; then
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/install-scripts?pullRequest=137) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/install-scripts?pullRequest=137) before we can accept your contribution.<br/><br/>:x: ujpandey [sign now](https://cla.dotnetfoundation.org/dotnet/install-scripts?pullRequest=137)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/install-scripts?pullRequest=137) it.</sub>

danmosemsft:
@bozturkMSFT could you please review?

lewing (commented):
Thanks @ujpandey

This fix looks fine to me but I'll wait for @bozturkMSFT to approve.

Looking over the original patch it might also make sense try wget if it is available but curl has failed.

lewing (approved):
Actually since this needs two

src/dotnet-install.sh
```diff
@@ -905,10 +905,10 @@ install_dotnet() {
 
     # The download function will set variables $http_code and $download_error_msg in case of failure.
     download "$download_link" "$zip_path" 2>&1 || download_failed=true
-    primary_path_http_code="$http_code"; primary_path_download_error_msg="$download_error_msg"
 
     #  if the download fails, download the legacy_download_link
     if [ "$download_failed" = true ]; then
+        primary_path_http_code="$http_code"; primary_path_download_error_msg="$download_error_msg"
```
hoyosjs, line 911: This will still be unbound if you don't have wget or curl - download will fail, but you won't have an error code. @bozturkMSFT does it make sense to say, just exit the script if neither was found?

ujpandey: That sounds reasonable enough to me at least so I updated the PR to do exactly that.

src/dotnet-install.sh
```diff
@@ -753,9 +753,8 @@ download() {
         elif machine_has "wget"; then
             downloadwget "$remote_path" "$out_path" || failed=true
         else
-            unset http_code
-            download_error_msg="Missing dependency: neither curl nor wget was found."
-            break
+            say_err "Missing dependency: neither curl nor wget was found."
```
hoyosjs, line 756: I'd recommend that you return to the old logic here. There should be no `http_code` if there's no http request.

src/dotnet-install.sh
hoyosjs: Returning 1 from download() will sadly preserve this error behavior. Exiting directly from that function wont .

There's two options:
- Just log the message there and exit. There's little that can be done at any other point in the script that's useful to the users. This also avoids doing retries if there's no curl or wget.
- Change this to something like:
```suggestion
        primary_path_http_code="${http_code:-toolError}"; primary_path_download_error_msg="$download_error_msg"
```
The messages will repeat and do retries, but it's pretty small of a fix. 


Even with this contribution I'd be happy, as most error cases in the script will be "couldn't download", not "couldn't find something to download with", and this would fix issues that are pretty disruptive right now.

src/dotnet-install.sh
ujpandey: Do you mean that I should be unsetting it? We are exiting the script when we hit this condition though, so that doesn't matter unless I am missing something? `http_code` is only a global and not an env var.

FWIW it appears that the script was already exiting if neither curl or wget were present thanks to `check_min_reqs()`.

hoyosjs: Ah, then maybe this is defensive. return 1 wouldn't exit here, but if check_min_reqs does it, then your original fix was better. Just means that the no-wget-or-curl branch of this was dead all along. Thanks for explaining.

ujpandey: Yep, I missed that about returning 1 from inside the function not actually exiting, thank you for catching that. I think it makes sense to be defensive also even if the branch is dead, and I think it's better to just exit with the message logged so I updated the PR to that fwiw.

idle-code (approved):
LGTM + this fixes our pipeline

bozturkMSFT (approved):

vlada-shubina (approved):
nit: it would be good to avoid whitespace changes when possible

dave-yotta:
Just to reiterate this has stopped all our deployments and a lot of development work, we rely on roslyn in production, please can the fix be merged ASAP or the latest install scripts be rolled back to yesterday.

Issue #125, opened 2021-01-04
Install scripts may fail when TERM is not set.


tmds:
I ran the install script in a `centos:8` container which doesn't have the TERM envvar set and it failed with:

```
tput: No value for $TERM and no -T specified
Failed to install dotnet SDK from public location (exit code '2').
Failed to install dotnet SDK from public location (exit code '2').
```

bozturkMSFT:
Hey @tmds,
 Can you let me know how you are impacted by this? Is this an ongoing incident? Can you workaround by setting the TERM yourself?

tmds:
> Can you workaround by setting the TERM yourself?

Yes.

bozturkMSFT:
Hi @tmds,
 I am glad to hear that you have a workaround to unblock you. I will close this since it seems to be a system configuration issue rather than an install script issue.
 Let us know if you need further help with this.

bozturkMSFT:
This was fixed in https://github.com/dotnet/install-scripts/pull/154 and will be available in the website after the deployment.

Issue #153, opened 2021-02-23
Installing from an an active channel with version latest leads to sporadic failures


nkolev92:
On our CI we try to install the latest 5.0.2xx SDK to test against. Unfortunately this leads to sporadic failures, and does not consistently work. 

```
[15:29:06 +0]   C:\Users\Roki2\Documents\Code\NuGet\NuGet.Client\cli\dotnet-install.ps1 -Channel release/5.0.2xx -i C:\Users\Roki2\Documents\Code\NuGet\NuGet.Client\cli -Version latest -Architecture x64 -NoPath 
dotnet-install: Note that the intended use of this script is for Continuous Integration (CI) scenarios, where: 
dotnet-install: - The SDK needs to be installed without user interaction and without admin rights. 
dotnet-install: - The SDK installation doesn't need to persist across multiple CI runs. 
dotnet-install: To set up a development environment or to run apps, use installers rather than this script. Visit https://dotnet.microsoft.com/download to get the installer.  
dotnet-install: Downloading primary link https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201-servicing.21123.3/dotnet-sdk-5.0.201-servicing.21123.3-win-x64.zip 
dotnet-install: The resource at https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201-servicing.21123.3/dotnet-sdk-5.0.201-servicing.21123.3-win-x64.zip is not available. 
dotnet-install: Downloading legacy link https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201-servicing.21123.3/dotnet-dev-win-x64.5.0.201-servicing.21123.3.zip 
dotnet-install: The resource at https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201-servicing.21123.3/dotnet-dev-win-x64.5.0.201-servicing.21123.3.zip is not available.
```

bozturkMSFT:
Hi @rbhanda,
 Latest version file for the branch `release/5.0.2xx` is at `https://dotnetcli.blob.core.windows.net/dotnet/Sdk/release/5.0.2xx/latest.version` which contains

```
ff3eefe79ed466e2983b87ce305c1b045294219b
5.0.201-servicing.21123.3
```

[...]

PR #154, opened 2021-02-24
Harden tput command against unset TERM

akoeplinger:
This was reported in https://github.com/dotnet/install-scripts/issues/125 and I just had another report of this in a different context.
There's no reason why we should fail to install dotnet just because we couldn't get the colors, so harden against `tput` failing.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -24,7 +24,7 @@ exec 3>&1
 # See if stdout is a terminal
 if [ -t 1 ] && command -v tput > /dev/null; then
     # see if it supports colors
-    ncolors=$(tput colors)
+    ncolors=$(tput colors || echo 0)
     if [ -n "$ncolors" ] && [ $ncolors -ge 8 ]; then
         bold="$(tput bold       || echo)"
         normal="$(tput sgr0     || echo)"


akoeplinger:
note: you can easily test this by running `TERM= ./dotnet-install.sh`.

akoeplinger:
Looks like downloading 5.0.2xx SDKs is broken, happens on master too:

```
$ ./dotnet-install.sh -i foo --channel "release/5.0.2xx"

dotnet-install: Downloading primary link https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201-servicing.21123.3/dotnet-sdk-5.0.201-servicing.21123.3-osx-x64.tar.gz
curl: (22) The requested URL returned error: 404
dotnet-install: The resource at https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201-servicing.21123.3/dotnet-sdk-5.0.201-servicing.21123.3-osx-x64.tar.gz is not available.
dotnet-install: Downloading legacy link https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201-servicing.21123.3/dotnet-dev-osx-x64.5.0.201-servicing.21123.3.tar.gz
curl: (22) The requested URL returned error: 404
dotnet-install: The resource at https://dotnetcli.azureedge.net/dotnet/Sdk/5.0.201-servicing.21123.3/dotnet-dev-osx-x64.5.0.201-servicing.21123.3.tar.gz is not available.
dotnet_install: Error: Could not find `.NET Core SDK` with version = 5.0.201-servicing.21123.3
dotnet_install: Error: Refer to: https://aka.ms/dotnet-os-lifecycle for information on .NET Core support
```

It's presumably the same issue as https://github.com/dotnet/install-scripts/issues/153, probably the SDK not yet published to the dotnetcli storage container?

bozturkMSFT:
@akoeplinger Thank you for the PR.
I will check the issue with 5.0.2xx, which will hopefully unblock this.

donJoseLuis (approved):

AR-May (approved):

PR #166, opened 2021-03-17
Rename usages of `Master` to `Main` in repo

DavidKarlas:


src/dotnet-install.ps1
```diff
@@ -17,7 +17,7 @@
     - 2-part version in a format A.B - represents a specific release
           examples: 2.0, 1.0
     - Branch name
-          examples: release/2.0.0, Master
+          examples: release/2.0.0, Main
```
bozturkMSFT, line 20: @mmitche Are we going to start publishing bits into main folder in blob storage now? 

This url is not pointing to anything yet: https://dotnetcli.blob.core.windows.net/dotnet/Sdk/main/latest.version

bozturkMSFT: @DavidKarlas I think we should revert this line since it doesn't work yet and there is no more time left to delay m2m transition.

bozturkMSFT (approved):

src/dotnet-install.ps1
```diff
@@ -17,7 +17,7 @@
     - 2-part version in a format A.B - represents a specific release
           examples: 2.0, 1.0
     - Branch name
-          examples: release/2.0.0, Master
+          example: release/2.0.0
```
AR-May, line 20: Shouldn't it be "examples: release/2.0.0, Master"?

AR-May (approved):

src/dotnet-install.ps1
AR-May: *it is fine, the decision was to remove "Master" as an example here.

PR #202, opened 2021-08-03
Disable retries on 404

### LINKED ISSUE #200: When dotnet is already installed the script still retries multiple times
Have a script that installs dotnet, and once installed run it again. The logic that queries for the vesion txt will retry 3 times with 1 second delay between checks because it considers 404 to be a failure, instead of correctly identifying that the resource does not exist.

This makes our script that installs 8 runtimes very slow even when we have the runtimes already installed.

```powershell

$sb = [ScriptBlock]::Create((Invoke-WebRequest -Uri https://raw.githubusercontent.com/dotnet/install-scripts/main/src/dotnet-install.ps1 -UseBasicParsing).Content)
& $sb -InstallDir "$PSScriptRoot/dotnet" -Runtime 'dotnet' -Version '2.1.0' -Channel '2.1.0' -NoPath -Verbose
``` 

```log
VERBOSE: dotnet-install: Normalized product: 'dotnet-runtime'
VERBOSE: dotnet-install: Get-Specific-Version-From-Version -AzureFeed "https://dotnetcli.azureedge.net/dotnet" -Channel "2.1.0" -Version "2.1.0" -JSonFile ""
VERBOSE: dotnet-install: Get-Download-Link -AzureFeed "https://dotnetcli.azureedge.net/dotnet" -SpecificVersion "2.1.0" -CLIArchitecture "x64"
VERBOSE: dotnet-install: Get-Product-Version -AzureFeed "https://dotnetcli.azureedge.net/dotnet" -SpecificVersion "2.1.0"
VERBOSE: dotnet-install: Get-Product-Version-Url -Flattened "True" -AzureFeed "https://dotnetcli.azureedge.net/dotnet" -SpecificVersion "2.1.0" -PackageDownloadLink ""
VERBOSE: dotnet-install: Constructed productVersion link: https://dotnetcli.azureedge.net/dotnet/Runtime/2.1.0/runtime-productVersion.txt
VERBOSE: dotnet-install: Get-Product-Version-Url -Flattened "False" -AzureFeed "https://dotnetcli.azureedge.net/dotnet" -SpecificVersion "2.1.0" -PackageDownloadLink ""

VERBOSE: dotnet-install: Constructed productVersion link: https://dotnetcli.azureedge.net/dotnet/Runtime/2.1.0/productVersion.txt
VERBOSE: dotnet-install: Checking for the existence of https://dotnetcli.azureedge.net/dotnet/Runtime/2.1.0/runtime-productVersion.txt
VERBOSE: dotnet-install: Exception ignored: The property 'OriginalString' cannot be found on this object. Verify that the property exists..Exception.Message - moving forward...
VERBOSE: dotnet-install: Exception ignored: The property 'OriginalString' cannot be found on this object. Verify that the property exists..Exception.Message - moving forward...
VERBOSE: dotnet-install: Exception ignored: The property 'OriginalString' cannot be found on this object. Verify that the property exists..Exception.Message - moving forward...
VERBOSE: dotnet-install: Could not read productVersion.txt at https://dotnetcli.azureedge.net/dotnet/Runtime/2.1.0/runtime-productVersion.txt (Exception: 'Unable to download https://dotnetcli.azureedge.net/dotnet/Runtime/2.1.0/runtime-productVersion.txt.'. )

VERBOSE: dotnet-install: Checking for the existence of https://dotnetcli.azureedge.net/dotnet/Runtime/2.1.0/productVersion.txt
VERBOSE: dotnet-install: Exception ignored: The property 'OriginalString' cannot be found on this object. Verify that the property exists..Exception.Message - moving forward...
VERBOSE: dotnet-install: Exception ignored: The property 'OriginalString' cannot be found on this object. Verify that the property exists..Exception.Message - moving forward...
VERBOSE: dotnet-install: Exception ignored: The property 'OriginalString' cannot be found on this object. Verify that the property exists..Exception.Message - moving forward...
VERBOSE: dotnet-install: Could not read productVersion.txt at https://dotnetcli.azureedge.net/dotnet/Runtime/2.1.0/productVersion.txt (Exception: 'Unable to download https://dotnetcli.azureedge.net/dotnet/Runtime/2.1.0/productVersion.txt.'. )

VERBOSE: dotnet-install: Using the default value '2.1.0' as the product version.
VERBOSE: dotnet-install: Constructed primary named payload URL: https://dotnetcli.azureedge.net/dotnet/Runtime/2.1.0/dotnet-runtime-2.1.0-win-x64.zip
VERBOSE: dotnet-install: Get-LegacyDownload-Link -AzureFeed "https://dotnetcli.azureedge.net/dotnet" -SpecificVersion "2.1.0" -CLIArchitecture "x64"
VERBOSE: dotnet-install: Constructed legacy named payload URL: https://dotnetcli.azureedge.net/dotnet/Runtime/2.1.0/dotnet-win-x64.2.1.0.zip
VERBOSE: dotnet-install: Resolve-Installation-Path -InstallDir "C:\Users\jajares\Desktop/dotnet"
```

https://github.com/dotnet/install-scripts/blob/e259f52ba2adb56da4038ac3c7c58c2e28b395cf/src/dotnet-install.ps1#L305

bekir-ozturk:
Fixes #200

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -190,7 +190,7 @@ function Invoke-With-Retry([ScriptBlock]$ScriptBlock, [System.Threading.Cancella
         }
         catch {
             $Attempts++
-            if (($Attempts -lt $MaxAttempts) -and (True -ne $cancellationToken.IsCancellationRequested)) {
+            if (($Attempts -lt $MaxAttempts) -and -not $cancellationToken.IsCancellationRequested) {
                 Start-Sleep $SecondsBetweenAttempts
             }
             else {
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -181,7 +181,7 @@ function Say-Invocation($Invocation) {
     Say-Verbose "$command $args"
 }
 
-function Invoke-With-Retry([ScriptBlock]$ScriptBlock, [int]$MaxAttempts = 3, [int]$SecondsBetweenAttempts = 1) {
+function Invoke-With-Retry([ScriptBlock]$ScriptBlock, [System.Threading.CancellationToken]$cancellationToken = [System.Threading.CancellationToken]::None, [int]$MaxAttempts = 3, [int]$SecondsBetweenAttempts = 1) {
     $Attempts = 0
 
     while ($true) {
@@ -190,7 +190,7 @@ function Invoke-With-Retry([ScriptBlock]$ScriptBlock, [int]$MaxAttempts = 3, [in
         }
         catch {
             $Attempts++
-            if ($Attempts -lt $MaxAttempts) {
+            if (($Attempts -lt $MaxAttempts) -and (True -ne $cancellationToken.IsCancellationRequested)) {
                 Start-Sleep $SecondsBetweenAttempts
             }
             else {
@@ -306,8 +306,9 @@ function Load-Assembly([string] $Assembly) {
 
 function GetHTTPResponse([Uri] $Uri, [bool]$HeaderOnly, [bool]$DisableRedirect, [bool]$DisableFeedCredential)
 {
-    Invoke-With-Retry(
-    {
+    $cts = New-Object System.Threading.CancellationTokenSource
+
+    $downloadScript = {
 
         $HttpClient = $null
 
@@ -373,6 +374,11 @@ function GetHTTPResponse([Uri] $Uri, [bool]$HeaderOnly, [bool]$DisableRedirect, 
                 if ($null -ne $Response) {
                     $DownloadException.Data["StatusCode"] = [int] $Response.StatusCode
                     $DownloadException.Data["ErrorMessage"] = "Unable to download $Uri. Returned HTTP status code: " + $DownloadException.Data["StatusCode"]
+
+                    if (404 -eq [int] $Response.StatusCode)
+                    {
+                        $cts.Cancel()
+                    }
                 }
 
                 throw $DownloadException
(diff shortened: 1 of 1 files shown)


src/dotnet-install.ps1
```diff
@@ -190,7 +190,7 @@ function Invoke-With-Retry([ScriptBlock]$ScriptBlock, [int]$MaxAttempts = 3, [in
         }
         catch {
             $Attempts++
-            if ($Attempts -lt $MaxAttempts) {
+            if (($Attempts -lt $MaxAttempts) -and (True -ne $cancellationToken.IsCancellationRequested)) {
```
nohwnd, line 193: This won't work, True is a function here, and you don't define it.
```suggestion
            if (($Attempts -lt $MaxAttempts) -and -not $cancellationToken.IsCancellationRequested) {
```

nohwnd (approved):
This is one way to do this. But it feels like a hack, because it adds to the jungle of how a request is processed in your code. 

I would much rather see that it would not retry on 404 directly, or that you could specify when it should not retry rather than "cancelling" and operation.

But if you are after the lowest hanging fruit, then it is okay I guess.

AR-May:
I know that in theory 404 error should mean that we are asking for a wrong URI and retries would not help with non-existent page, therefore we can skip them. I have some worries though, as I was in situation that simple refresh helped with 404 error on a web page, so it was a temporary problem which resolved with retries.  
So, I am a bit reluctant to do such changes without additional thinking on this topic. What if we break good cases for  retry on 404, when it helps? Do we have them, is it possible somehow to check that?

nohwnd:
@AR-May I agree that it can happen that the server returns 404 and on the next try the resource appears, but that is not standard. 4xx range of errors means the user did something wrong, and I don't think we should retry on them, because of a slim chance that the resource would appear. 

If you are concerned about reliability then load balancing should be used instead, and that should return 5xx range errors so we know to that we should retry.

 Ideally, you'd check if the version given is very specific, if not you'd check the server what the latest version is, no retry, then check my folder if I have it, and if not you'd try to reliably download it. But imho the script is now too complex to easily capture such scenario and unfortunately you don't have time to re-write it.

AR-May (approved):
I feel indeed that we are addressing the problem not quite in a proper way. 

I also agree that server should return another error in this case. So, let's hope it is the case, as we should assume by default that server behaves up to standards. 
(But if we run into this problem, it seems not quite resolvable from our side. And I suppose in this case we would need to revert. We also should be careful about that when deploy.)

Also, I think we'd better to keep logic for sh and ps1 scripts as similar as possible, so we'd better skip retries for 404 there too. But that is up to you, @bekir-ozturk. 

Well, as a quick fix this might be OK, I suppose. We can try it and see, if there would be reports of getting 404 error which are temporary.

bekir-ozturk:
I agree that it would be the right move to implement the same logic in .sh script as well. I will merge this as is now due to the availability limitations and leave improvements for later.

PR #208, opened 2021-09-03
Fix: machine_has is not working reliably on Alpine

### LINKED ISSUE #206: Script unable to find installed curl / wget on Alpine
```
bash-5.1# chmod +x dotnet-install.sh
bash-5.1# ./dotnet-install.sh -c Current --os linux-musl

dotnet_install: Error: curl (recommended) or wget are required to download dotnet. Install missing prerequisite to proceed.
bash-5.1# which curl
/usr/bin/curl
bash-5.1# which wget
/usr/bin/wget
bash-5.1# cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.14.0
PRETTY_NAME="Alpine Linux v3.14"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"
bash-5.1# 
```

bekir-ozturk:
Fixes #206

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -224,7 +224,7 @@ get_legacy_os_name() {
 machine_has() {
     eval $invocation
 
-    hash "$1" > /dev/null 2>&1
+    command -v "$1" > /dev/null 2>&1
     return $?
 }
 


AR-May (approved):
LGTM

hoyosjs (approved):

vlada-shubina (approved):

PR #211, opened 2021-09-14
Install on non-windows systems using dotnet-install.ps1

nohwnd:
Allow installation on non-windows systems using PowerShell. PowerShell is commonly installed on CI system (e.g. AzDO, GitHub) and this avoids the need to switch installation approaches based on operating system.

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -227,12 +227,12 @@ function Get-Machine-Architecture() {
     # PS x64 doesn't define this, so we fall back to PROCESSOR_ARCHITECTURE.
     # Possible values: amd64, x64, x86, arm64, arm
 
-    if( $ENV:PROCESSOR_ARCHITEW6432 -ne $null )
+    if( $env:PROCESSOR_ARCHITEW6432 -ne $null )
     {    
-        return $ENV:PROCESSOR_ARCHITEW6432
+        return $env:PROCESSOR_ARCHITEW6432
     }
 
-    return $ENV:PROCESSOR_ARCHITECTURE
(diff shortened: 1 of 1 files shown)


src/dotnet-install.ps1
```diff
@@ -275,6 +294,19 @@ function Get-NormalizedProduct([string]$Runtime) {
     }
 }
 
+function Get-NormalizedOs([string]$Os) {
+    Say-Invocation $MyInvocation
+
+    # REVIEW: I know I am missing a lot here.
```
nohwnd, line 300: I am aware that I am missing a lot of the logic for different systems here, it should be simple porting it from the .sh script, but I want to first know if you are willing to take this change before committing deeper.

bekir-ozturk:
Hi @nohwnd,
 We discussed this internally as a team and decided to not take the change for the following reasons:
 - We believe this change brings very little benefit. Bash version of the script already delivers all the needs in Linux area. PowerShell script doesn't bring any features that isn't already available via Bash.
 - The use of this wouldn't be straightforward as it requires having PowerShell Core installed on the machine.
 - Having this scenario supported will require commitment in the long run. Once announced, we cannot go back as this would mean breaking any users that are already using Ps1 script on Linux. Unfortunately, our team doesn't have the availability to commit to a new feature such as this one at the moment.

nohwnd:
Thanks for the feedback, but both point 1 and 2 are the reason that this change was created in the first place. Both GH Actions, and AzDo have PowerShell installed and ready to go. This change would allow you to install dotnet on both Windows and non-Windows using **the same** script.

I agree with 3, it is unfortunate that more time cannot be dedicated to this.

PR #213, opened 2021-09-15
Don't use unavailable wget options

bekir-ozturk:
Fixes https://github.com/dotnet/install-scripts/issues/216

**Problem:** Some options of `wget` are not available on Alpine.

**Solution:** The changes in this PR detects the missing options and removes them.
The implication of this removal will be that scripts running on alpine+wget will be quicker to give up on retrying. This should be acceptable considering that:
- We already have a custom retry logic built-in in a higher level,
- The issue can simply be resolved by installing curl.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -642,7 +642,7 @@ get_specific_product_version() {
             fi
         elif machine_has "wget"
         then
-            specific_product_version=$(wget -qO- "${download_link}${feed_credential}")
+            specific_product_version=$(wget -qO- "${download_link}${feed_credential}" 2>&1)
             if [ $? = 0 ]; then
                 echo "${specific_product_version//[$'\t\r\n']}"
                 return 0
@@ -918,6 +918,8 @@ get_http_header_wget() {
     eval $invocation
     local remote_path="$1"
     local disable_feed_credential="$2"
+    local wget_output=''
+    local wget_result=''
 
     remote_path_with_credential="$remote_path"
     if [ "$disable_feed_credential" = false ]; then
@@ -925,8 +927,17 @@ get_http_header_wget() {
     fi
 
     wget_options="-q -S --spider --tries 5 --waitretry 2 --connect-timeout 15 "
-    wget $wget_options "$remote_path_with_credential" 2>&1 || return 1
-    return 0
+    wget_output=$(wget $wget_options "$remote_path_with_credential" 2>&1)
+    wget_result=$?
+
+    if [[ $wget_result != 0 ]] && [[ $wget_output == *"unrecognized option:"* ]]; then
+        wget_options="-q -S --spider --tries 5"
+        wget $wget_options "$remote_path_with_credential" 2>&1 || return 1
+        return 0
+    fi
+
+    echo $wget_output
+    return $wget_result
 }
 
 # args:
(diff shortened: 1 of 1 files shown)


src/dotnet-install.sh
```diff
@@ -918,15 +918,26 @@ get_http_header_wget() {
...
     fi
 
     wget_options="-q -S --spider --tries 5 --waitretry 2 --connect-timeout 15 "
-    wget $wget_options "$remote_path_with_credential" 2>&1 || return 1
-    return 0
+    wget_output=$(wget $wget_options "$remote_path_with_credential" 2>&1)
+    wget_result=$?
+
+    if [[ $wget_result != 0 ]] && [[ $wget_output == *"unrecognized option:"* ]]; then
+        wget_options="-q -S --spider --tries 5"
```
ladipro, line 934: super-nit: Worth keeping the base options in a separate variable to avoid duplicating them on line 929 and here?

src/dotnet-install.sh
```diff
@@ -918,15 +918,26 @@ get_http_header_wget() {
...
         remote_path_with_credential+="$feed_credential"
     fi
 
     wget_options="-q -S --spider --tries 5 --waitretry 2 --connect-timeout 15 "
-    wget $wget_options "$remote_path_with_credential" 2>&1 || return 1
-    return 0
+    wget_output=$(wget $wget_options "$remote_path_with_credential" 2>&1)
+    wget_result=$?
+
+    if [[ $wget_result != 0 ]] && [[ $wget_output == *"unrecognized option:"* ]]; then
```
ladipro, line 933: Can the system be localized and configured such that `wget` prints errors in another language?

src/dotnet-install.sh
bekir-ozturk: Should be done with the latest commit now.

src/dotnet-install.sh
bekir-ozturk: I removed the string comparison and now we rely on the exit status code.
2 means parsing error: https://www.gnu.org/software/wget/manual/wget.html#Exit-Status

src/dotnet-install.sh
```diff
@@ -918,15 +918,28 @@ get_http_header_wget() {
...
+    wget_output=$(wget $wget_options $wget_options_extra "$remote_path_with_credential" 2>&1)
+    wget_result=$?
+
+    if [[ $wget_result == 2 ]]; then
+        # Parsing of the command has failed. Exclude potentially unrecognized options and retry.
+        wget $wget_options "$remote_path_with_credential" 2>&1 || return 1
+        return 0
+    fi
+
+    echo $wget_output
```
ladipro, line 941: Are the new `echo`'s intentional?

ladipro (approved):
Looks good, thank you.

danmoseley:
Thank you. Do you plan to add regular testing on Alpine (probably the earliest version we support) to avoid this regressing?

bekir-ozturk:
> Thank you. Do you plan to add regular testing on Alpine (probably the earliest version we support) to avoid this regressing?

We started working on an improved monitoring solution that covered more OSes, but it was temporarily put on hold because of a shift in the team's priorities. We hope to get back to it once we have the availability.

DavidKarlas (approved):

src/dotnet-install.sh
bekir-ozturk: Removed

PR #260, opened 2022-03-03
fix bashscript for run in sudo sh environment

YuliiaKovalova:
fix bashscript for run in `sudo sh` environment.
Guideline for fix: [SC2181](https://github.com/koalaman/shellcheck/wiki/SC2181)

Install-scripts log with applied changes

```
ykovalova@NETUXs-MacBook-Pro ~ % sudo sh dotnet-install.sh --version "6.0.200" --install-dir "/Users/runner/hostedtoolcache888/dotnet" --verbose
dotnet-install: Note that the intended use of this script is for Continuous Integration (CI) scenarios, where:
dotnet-install: - The SDK needs to be installed without user interaction and without admin rights.
dotnet-install: - The SDK installation doesn't need to persist across multiple CI runs.
dotnet-install: To set up a development environment or to run apps, use installers rather than this script. Visit https://dotnet.microsoft.com/download to get the installer.

dotnet-install: Calling: machine_has curl
dotnet-install: Calling: calculate_vars
dotnet-install: Calling: get_normalized_architecture_from_architecture <auto>
dotnet-install: Calling: get_machine_architecture
dotnet-install: Normalized architecture: 'x64'.
dotnet-install: Calling: get_normalized_os
dotnet-install: Calling: get_current_os_name
dotnet-install: Normalized OS: 'osx'.
dotnet-install: Calling: get_normalized_quality
dotnet-install: Normalized quality: ''.
dotnet-install: Calling: get_normalized_channel LTS
dotnet-install: Normalized channel: 'LTS'.
dotnet-install: Calling: get_normalized_product
dotnet-install: Normalized product: 'dotnet-sdk'.
dotnet-install: Calling: resolve_installation_path /Users/runner/hostedtoolcache888/dotnet
dotnet-install: InstallRoot: '/Users/runner/hostedtoolcache888/dotnet'.
dotnet-install: Calling: get_specific_version_from_version https://dotnetcli.azureedge.net/dotnet LTS x64 6.0.200
dotnet-install: Calling: get_specific_product_version https://dotnetcli.azureedge.net/dotnet 6.0.200
dotnet-install: Calling: get_specific_product_version_url https://dotnetcli.azureedge.net/dotnet 6.0.200 true
dotnet-install: Constructed productVersion link: https://dotnetcli.azureedge.net/dotnet/Sdk/6.0.200/sdk-productVersion.txt
dotnet-install: Calling: get_specific_product_version_url https://dotnetcli.azureedge.net/dotnet 6.0.200 false
[...]

PR #261, opened 2022-03-03
Fix wget path on Alpine

### LINKED ISSUE #259: script does not work on Alpine due to failed fallback for unrecognized options
On Alpine, the BusyBox version of wget does not understand `--waitretry` (at least). This was reported in https://github.com/dotnet/install-scripts/issues/216 but the fix https://github.com/dotnet/install-scripts/pull/213 assumes that wget returns 2 for unexpected parameters. This is the case on Ubuntu but not on BusyBox:

```sh

# Ubuntu
dan@danmoseL:~$ uname -a
Linux danmoseL 5.10.60.1-microsoft-standard-WSL2 #1 SMP Wed Aug 25 23:20:18 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
dan@danmoseL:~$ cat /etc/os-release | grep PRETTY
PRETTY_NAME="Ubuntu 20.04.2 LTS"
dan@danmoseL:~$ wget -h  | head -1
GNU Wget 1.20.3, a non-interactive network retriever.
dan@danmoseL:~$ wget --foofoo
wget: unrecognized option '--foofoo'
Usage: wget [OPTION]... [URL]...

Try `wget --help' for more options.
dan@danmoseL:~$ echo $?
2
```
```sh
# Alpine
danmoseL:~/git/performance# uname -a
Linux danmoseL 5.10.60.1-microsoft-standard-WSL2 #1 SMP Wed Aug 25 23:20:18 UTC 2021 x86_64 Linux
danmoseL:~/git/performance# cat /etc/os-release | grep PRETTY
PRETTY_NAME="Alpine Linux v3.13"
danmoseL:~/git/performance# wget 2>&1 >/dev/null | head -1
BusyBox v1.32.1 () multi-call binary.
danmoseL:~/git/performance# wget https://google.com/ 2&>1 > nul
echodanmoseL:~/git/performance# echo $?
0
danmoseL:~/git/performance# wget --foofoo 2&>1 nul
danmoseL:~/git/performance# echo $?
1
```

So the original bug is not fixed for me and on a fresh Alpine WSL2 (which does not have curl either), install-script does not work.

This fixes it on my Alpine: https://github.com/dotnet/install-scripts/compare/main...danmoseley:fix.wget?expand=1

Before fix, it fails:
```sh
danmoseL:~/git/performance# /root/git/performance/tools/dotnet/x64/dotnet-install.sh -InstallDir /root/git/performance/tools/dotnet/x64 -Architecture x64 -Channel 6.0.1xx

dotnet-install: Download attempt #1 has failed:  Unable to download https://dotnetcli.azureedge.net/dotnet/Sdk/6.0.1xx/latest.version. Returned HTTP status code: .
dotnet-install: Attempt #2 will start in 10 seconds.
dotnet-install: Download attempt #2 has failed:  Unable to download https://dotnetcli.azureedge.net/dotnet/Sdk/6.0.1xx/latest.version. Returned HTTP status code: .
dotnet-install: Attempt #3 will start in 20 seconds.
dotnet-install: Download attempt #1 has failed:  Unable to download https://dotnetbuilds.azureedge.net/public/Sdk/6.0.1xx/latest.version. Returned HTTP status code: .
dotnet-install: Attempt #2 will start in 10 seconds.
dotnet-install: Download attempt #2 has failed:  Unable to download https://dotnetbuilds.azureedge.net/public/Sdk/6.0.1xx/latest.version. Returned HTTP status code: .
dotnet-install: Attempt #3 will start in 20 seconds.
dotnet_install: Error: Failed to resolve the exact version number.
danmoseL:~/git/performance#
```

After fix:
```sh
danmoseL:~/git/performance# /root/git/performance/tools/dotnet/x64/dotnet-install.sh -InstallDir /root/git/performance/tools/dotnet/x64 -Architecture x64 -Channel 6.0.1xx

dotnet-install: Attempting to download using aka.ms link https://dotnetcli.azureedge.net/dotnet/Sdk/6.0.102/dotnet-sdk-6.0.102-linux-musl-x64.tar.gz
wget: unrecognized option: waitretry
BusyBox v1.32.1 () multi-call binary.

Usage: wget [-c|--continue] [--spider] [-q|--quiet] [-O|--output-document FILE]
        [-o|--output-file FILE] [--header 'header: value'] [-Y|--proxy on/off]
        [-P DIR] [-S|--server-response] [-U|--user-agent AGENT] [-T SEC] URL...

Retrieve files via HTTP or FTP

        --spider        Only check URL existence: $? is 0 if exists
        -c              Continue retrieval of aborted transfer
        -q              Quiet
        -P DIR          Save to DIR (default .)
        -S              Show server response
        -T SEC          Network read timeout is SEC seconds
        -O FILE         Save to FILE ('-' for stdout)
        -o FILE         Log messages to FILE
        -U STR          Use STR for User-Agent header
        -Y on/off       Use proxy
Connecting to dotnetcli.azureedge.net (72.21.81.200:443)
saving to '/tmp/dotnet.XXXEAaDoD'
dotnet.XXXEAaDoD       0% |                                                              
```

It's unfortunate that the output contains the dump of wget options here, when it's working fine (this would already be happening on Ubuntu). Perhaps the simplest way to avoid this would be to scan `wget -h` for `--waitretry` and if it's NOT present, then go straight to the fallback call. I can add this to the PR if you like.

Are we testing this script regularly on Alpine -- if so perhaps on a system with a custom wget? Surely it is not all broken on Alpine?

cc @adamsitnik @jeffhandley

danmoseley:
Fix #259 

1. Test `wget -h` output for the two extra parameters we want to use, and only attempt to use them if we see them there. Pipe to /dev/null so that there's no visible output.
2. Leave the fallback path in to reduce risk. However BusyBox wget sets exit code 1 when there are unexpected parameters and GNU wget returns 2. Fix this path by checking for either.

Tested on Alpine and Ubuntu (forcing the wget path). Verified it works, and that on Ubuntu it does use the extra wget parameters.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -926,7 +926,7 @@ get_http_header_wget() {
     local wget_result=''
 
     # Test for options that aren't supported on all wget implementations.
-    wget -h 2>&1 | grep "xwaitretry" >null && wget -h 2>&1 | grep "connect-timeout" >null
+    wget -h 2>&1 | grep "waitretry" >/dev/null && wget -h 2>&1 | grep "connect-timeout" >/dev/null
     if [ $? = 0 ]; then
         wget_options_extra="--waitretry 2 --connect-timeout 15 "
     fi
@@ -1042,7 +1042,7 @@ downloadwget() {
     local wget_result=''
 
     # Test for options that aren't supported on all wget implementations.
-    wget -h 2>&1 | grep "xwaitretry" >null && wget -h 2>&1 | grep "connect-timeout" >null
+    wget -h 2>&1 | grep "waitretry" >/dev/null && wget -h 2>&1 | grep "connect-timeout" >/dev/null
     if [ $? = 0 ]; then
         wget_options_extra="--waitretry 2 --connect-timeout 15 "
     fi
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -921,9 +921,15 @@ get_http_header_wget() {
     local remote_path="$1"
     local disable_feed_credential="$2"
     local wget_options="-q -S --spider --tries 5 "
-    # Store options that aren't supported on all wget implementations separately.
-    local wget_options_extra="--waitretry 2 --connect-timeout 15 "
+
+    local wget_options_extra=''
     local wget_result=''
+
+    # Test for options that aren't supported on all wget implementations.
+    wget -h 2>&1 | grep "xwaitretry" >nul && wget -h 2>&1 | grep "connect-timeout" >nul
+    if [ $? = 0 ]; then
+        wget_options_extra="--waitretry 2 --connect-timeout 15 "
+    fi
 
     remote_path_with_credential="$remote_path"
     if [ "$disable_feed_credential" = false ]; then
@@ -933,7 +939,7 @@ get_http_header_wget() {
     wget $wget_options $wget_options_extra "$remote_path_with_credential" 2>&1
     wget_result=$?
 
-    if [[ $wget_result == 2  ]] || [[ $wget_result == 1 ]]; then
+    if [[ $wget_result == 1  ]] || [[ $wget_result == 2 ]]; then
         # Parsing of the command has failed. Exclude potentially unrecognized options and retry.
         say_verbose "wget parsing failed. trying again with fewer options.."
         wget $wget_options "$remote_path_with_credential" 2>&1
@@ -1031,9 +1037,15 @@ downloadwget() {
     # Append feed_credential as late as possible before calling wget to avoid logging feed_credential
     local remote_path_with_credential="${remote_path}${feed_credential}"
     local wget_options="--tries 20 "
-    # Store options that aren't supported on all wget implementations separately.
-    local wget_options_extra="--waitretry 2 --connect-timeout 15 "
+
+    local wget_options_extra=''
     local wget_result=''
+
+    # Test for options that aren't supported on all wget implementations.
+    wget -h 2>&1 | grep "xwaitretry" >nul && wget -h 2>&1 | grep "connect-timeout" >nul
+    if [ $? = 0 ]; then
+        wget_options_extra="--waitretry 2 --connect-timeout 15 "
+    fi
 
     if [ -z "$out_path" ]; then
         wget -q $wget_options $wget_options_extra -O - "$remote_path_with_credential" 2>&1
@@ -1043,7 +1055,7 @@ downloadwget() {
         wget_result=$?
     fi
 
-    if [[ $wget_result == 2  ]] || [[ $wget_result == 1 ]]; then
+    if [[ $wget_result == 1  ]] || [[ $wget_result == 2 ]]; then
         # Parsing of the command has failed. Exclude potentially unrecognized options and retry.
         say_verbose "wget parsing failed. trying again with fewer options.."
         if [ -z "$out_path" ]; then
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -933,8 +933,9 @@ get_http_header_wget() {
     wget $wget_options $wget_options_extra "$remote_path_with_credential" 2>&1
     wget_result=$?
 
-    if [[ $wget_result == 2 ]]; then
+    if [[ $wget_result == 2  ]] || [[ $wget_result == 1 ]]; then
         # Parsing of the command has failed. Exclude potentially unrecognized options and retry.
+        say_verbose "wget parsing failed. trying again with fewer options.."
         wget $wget_options "$remote_path_with_credential" 2>&1
         return $?
     fi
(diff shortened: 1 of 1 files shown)


danmoseley:
I'd be grateful if you'd do any testing necessary beyond what I did, as I'm rusty at shell scripts and I don't have all the distros to test, nor do I know this code.

YuliiaKovalova:
no objections from my side.
@bekir-ozturk  @vlada-shubina @AR-May please take a look too

AR-May (approved):
LGTM

src/dotnet-install.sh
```diff
@@ -921,10 +921,16 @@ get_http_header_wget() {
...
     local wget_options="-q -S --spider --tries 5 "
-    # Store options that aren't supported on all wget implementations separately.
-    local wget_options_extra="--waitretry 2 --connect-timeout 15 "
+
+    local wget_options_extra=''
     local wget_result=''
 
+    # Test for options that aren't supported on all wget implementations.
+    wget -h 2>&1 | grep "waitretry" >/dev/null && wget -h 2>&1 | grep "connect-timeout" >/dev/null
+    if [ $? = 0 ]; then
```
bekir-ozturk, line 930: We have recently fixed an issue with checking the results with `$?` here https://github.com/dotnet/install-scripts/pull/260.

Can we instead do something like:
```
if ! wget -h 2>&1 | grep .... ; then
```
instead. This will likely fail with `sudo` otherwise.

src/dotnet-install.sh
```diff
@@ -933,8 +939,9 @@ get_http_header_wget() {
     wget $wget_options $wget_options_extra "$remote_path_with_credential" 2>&1
     wget_result=$?
 
-    if [[ $wget_result == 2 ]]; then
+    if [[ $wget_result == 1 ]] || [[ $wget_result == 2 ]]; then
         # Parsing of the command has failed. Exclude potentially unrecognized options and retry.
+        say_verbose "wget parsing failed. trying again with fewer options.."
         wget $wget_options "$remote_path_with_credential" 2>&1
         return $?
     fi
```
bekir-ozturk, line 947: Now we have a better understanding of what arguments we can pass to the command at runtime.
I suggest we remove the retries.

bekir-ozturk (changes requested):

src/dotnet-install.sh
YuliiaKovalova: @danmoseley,
I will take it from here.

Thank you for fixing the main issue.

src/dotnet-install.sh
YuliiaKovalova: done

src/dotnet-install.sh
YuliiaKovalova: @bekir-ozturk ,

the fix is ready, I have tested it.
Please review

src/dotnet-install.sh
```diff
@@ -921,25 +921,24 @@ get_http_header_wget() {
...
     local wget_options="-q -S --spider --tries 5 "
-    # Store options that aren't supported on all wget implementations separately.
-    local wget_options_extra="--waitretry 2 --connect-timeout 15 "
-    local wget_result=''
+
+    local wget_options_extra=''
+
+    # Test for options that aren't supported on all wget implementations.
+    if ! wget -h 2>&1 | grep "waitretry" >/dev/null && wget -h 2>&1 | grep "connect-timeout" >/dev/null; then
+        say 'wget extra options are unavaiable for this environment'
```
danmoseley, line 929:
```suggestion
        say 'wget extra options are unavailable for this environment'
```

src/dotnet-install.sh
```diff
@@ -921,25 +921,24 @@ get_http_header_wget() {
...
+    else
+        wget_options_extra="--waitretry 2 --connect-timeout 15 "
+    fi
 
     remote_path_with_credential="$remote_path"
     if [ "$disable_feed_credential" = false ]; then
         remote_path_with_credential+="$feed_credential"
     fi
 
     wget $wget_options $wget_options_extra "$remote_path_with_credential" 2>&1
```
bekir-ozturk, line 939: Since we are getting rid of usages of `$?`, I think it may be better to do:
```suggestion
    return wget $wget_options $wget_options_extra "$remote_path_with_credential" 2>&1
```

bekir-ozturk (changes requested):

src/dotnet-install.sh
YuliiaKovalova: it produces an error
image

src/dotnet-install.sh
```diff
@@ -1042,8 +1047,9 @@ downloadwget() {
         wget_result=$?
     fi
 
-    if [[ $wget_result == 2 ]]; then
+    if [[ $wget_result == 1 ]] || [[ $wget_result == 2 ]]; then
         # Parsing of the command has failed. Exclude potentially unrecognized options and retry.
+        say_verbose "wget parsing failed. trying again with fewer options.."
         if [ -z "$out_path" ]; then
             wget -q $wget_options -O - "$remote_path_with_credential" 2>&1
```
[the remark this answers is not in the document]
YuliiaKovalova: done

src/dotnet-install.sh
bekir-ozturk: Does this work with `sudo` without the suggested fix? If yes, let's keep as is.

bekir-ozturk (approved):

PR #267, opened 2022-04-14
remove outdated dependency

YuliiaKovalova:
remove deprecated dependency according to https://dnceng.visualstudio.com/internal/_componentGovernance/dotnet-install-scripts/alert/6567780?typeId=6427340


bekir-ozturk (approved):
None of the Monitoring projects should be in use today. I suggest we remove them all.

PR #270, opened 2022-04-22
Progress for download

nohwnd:
Add progress to download using the standard PowerShell progress reporting.

progress

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -361,7 +361,10 @@ function GetHTTPResponse([Uri] $Uri, [bool]$HeaderOnly, [bool]$DisableRedirect, 
             # Defaulting to 20 minutes allows it to work over much slower connections.
             $HttpClient.Timeout = New-TimeSpan -Seconds $DownloadTimeout
 
-            if ($HeaderOnly){
+            $isZip = $Uri -like "*.zip"
+
+            # Zips are big, we want to report progress on them, so we grab headers first, and then we read the stream while reporting progress.
+            if ($HeaderOnly -or $isZip){
                 $completionOption = [System.Net.Http.HttpCompletionOption]::ResponseHeadersRead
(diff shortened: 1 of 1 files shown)


nohwnd:
Gauging your interest in merging this, in relation to #196 and https://github.com/dotnet/arcade/pull/7910. The implementation is not complete because later the contents are copied from the stream which is now already closed, and the memory stream (or similar) needs to be used instead, but it is 95% there I guess.

@kant2002 please give it a try if this looks interesting to you. My connection is fast, so I don't know if this would help.

kant2002:
@nohwnd I replace dotnet-install.ps1 in my local copy of runtime repository, and unfortunately issue still the same. Also please note that in Arcade there additional task https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/src/InstallDotNetCore.cs which probably would not display any progress which make this change somewhat irrelevant for me :face_exhaling:

nohwnd:
@YuliiaKovalova Yeah I am not surprised the test work, I did not finish the implementation past showing the progress. So when the .Content is read, it is trying to read from an already closed stream, and it throws. I did few changes in the past here, that were not merged, so I did just the minimal amount of work to show what I have in mind and that it is possible. With the complete implementation there should be neither of the issues you mention. 

@kant2002 ok. Good to know.

nohwnd:
@YuliiaKovalova So I guess I should go ahead with the implementation, and fix my issues so you can merge it? 🙂

YuliiaKovalova:
Hi @nohwnd,

No objections from my side, but let me prove this moment with the product owner and I will return to you as soon as possible.

YuliiaKovalova:
Hi @nohwnd,

Heads-up- we have a green light!

YuliiaKovalova:
close as not planned

Issue #277, opened 2022-05-23
Add code response for requests with timeout interruption


YuliiaKovalova:
Based on the feedback ticket [https://github.com/dotnet/arcade/issues/9156#issuecomment-1132188540](https://github.com/dotnet/arcade/issues/9156#issuecomment-1132188540), it is necessary to improve logging for timeout case.
Before doing the work please pick the exit code number according to [Advanced Bash-Scripting Guide](https://tldp.org/LDP/abs/html/exitcodes.html) and [BSD semi-standards](https://www.apt-browse.org/browse/ubuntu/trusty/main/amd64/libc6-dev/2.19-0ubuntu6/file/usr/include/sysexits.h).

Issue #247, opened 2022-01-21
Using Quality and Version together should give better error


bekir-ozturk:
Quality and Version options cannot be used together. If attempted, it results in the following error message:
```
Failed to resolve the exact version number.
```

We should clarify the message to better explain what went wrong.

YuliiaKovalova:
The fix will be included into the next deployment

PR #278, opened 2022-05-31
script enhancements

YuliiaKovalova:
- edit error message when quality and version both specified [issue 247](https://github.com/dotnet/install-scripts/issues/247)
- extend successful installation message with version info  [issue 268](https://github.com/dotnet/install-scripts/issues/268)
- fix small grammar mistakes (install scripts dev doc)
- fix the output message for a download timeout error [issue 277](https://github.com/dotnet/install-scripts/issues/277)


src/dotnet-install.ps1
```diff
@@ -244,6 +244,10 @@ function Get-NormalizedQuality([string]$Quality) {
         return ""
     }
 
+    if (![string]::IsNullOrEmpty($Version)) {
+        throw "Either Quality or Version option has to be specified. See https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script#:~:text=valid%20for%20Windows.)-,%2DQuality%7C%2D%2Dquality%20%3CQUALITY%3E,-Downloads%20the%20latest for details."
```
baronfel: This is useful, but I don't think this anchor-syntax is available on all browsers - I believe it is only supported on chromium-based browsers.  Is there a 'normal' HTML anchor on this docs page that can be used instead? If not, can we modify/work with the docs team to add something more linkable?

src/dotnet-install.sh
```diff
@@ -371,11 +371,19 @@ get_normalized_os() {
...
 get_normalized_quality() {
     eval $invocation
-
+    
     local quality="$(to_lowercase "$1")"
+    local version="$(to_lowercase "$2")"
+
     if [ ! -z "$quality" ]; then
+        if [ ! -z "$version" ] ; then
+            say_err "Either Quality or Version option has to be specified. See https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script#:~:text=valid%20for%20Windows.)-,%2DQuality%7C%2D%2Dquality%20%3CQUALITY%3E,-Downloads%20the%20latest for details."
```
baronfel: Similar question here as for the above.

src/dotnet-install.sh
```diff
@@ -645,11 +636,13 @@ get_specific_product_version() {
 
         if machine_has "curl"
         then
-            specific_product_version=$(curl -s --fail "${download_link}${feed_credential}" 2>&1)
-            if [ $? = 0 ]; then
+            if ! specific_product_version=$(curl -s --fail "${download_link}${feed_credential}" 2>&1); then
```
baronfel, line 640: good catch, this fixes a shellcheck warning 👍

baronfel (commented):
For the timeout exit code work, what is the error code that's returned? Do you have sample output from testing that I can see?  What I'm curious about is if the user has an exit-code-based way of detecting a script timeout error vs. any other kind of error.

src/dotnet-install.sh
```diff
@@ -1001,18 +1001,34 @@ downloadcurl() {
     local remote_path_with_credential="${remote_path}${feed_credential}"
     local curl_options="--retry 20 --retry-delay 2 --connect-timeout 15 -sSL -f --create-dirs "
     local failed=false
+    local curl_exit_code=0;
     if [ -z "$out_path" ]; then
-        curl $curl_options "$remote_path_with_credential" 2>&1 || failed=true
+        curl $curl_options "$remote_path_with_credential" 2>&1
+        curl_exit_code=$?
+        if [ $curl_exit_code -gt 0 ]; then
+            failed=true
```
bekir-ozturk, line 1009: We don't seem to need this `failed` variable anymore. It is used only once and we can just use `curl_exit_code -gt 0` instead.

src/dotnet-install.sh
```diff
@@ -1001,18 +1001,34 @@ downloadcurl() {
...
-        local response=$(get_http_header_curl $remote_path $disable_feed_credential)
-        http_code=$( echo "$response" | awk '/^HTTP/{print $2}' | tail -1 )
+    
+    if [[ "$failed" = true ]]; then
         download_error_msg="Unable to download $remote_path."
-        if  [[ $http_code != 2* ]]; then
+        # Check for curl timeout codes
+        if [[ $curl_exit_code == 7 || $curl_exit_code == 28 ]]; then
+            http_code="408"
             download_error_msg+=" Returned HTTP status code: $http_code."
```
bekir-ozturk, line 1024: I wonder if this will cause any confusion when diagnosing issues. After all, `curl(7)` and `curl(28)` are not the same as `http(408)`.

It is my understanding that timeout issues from curl mean that it has failed to reach the server (which points to a network issue) while http 408 means that the server has knowingly terminated the connection (which means network is fine).

Therefore, I suggest we leave the `http_code` untouched while updating the error msg
```suggestion
            download_error_msg+=" Failed to reach the server: connection timeout."
```

bekir-ozturk: Same concerns for `wget` error code `4`.

YuliiaKovalova: Agree!

src/dotnet-install.ps1
```diff
@@ -1159,6 +1164,11 @@ if ([string]::IsNullOrEmpty($NormalizedQuality) -and 0 -eq $DownloadLinks.count)
 }
 
 if ($DownloadLinks.count -eq 0) {
+
+    if ($Version.ToLowerInvariant() -ne "latest" -and -not [string]::IsNullOrEmpty($Quality)) {
```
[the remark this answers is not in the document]
YuliiaKovalova: It was a wrong place, just moved above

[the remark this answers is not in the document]
bekir-ozturk: Should we port the same thing to bash script as well?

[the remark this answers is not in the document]

bekir-ozturk (approved):

Issue #283, opened 2022-06-16
dotnet-install.ps1 when ran on windows-arm64 installs x64 version if --architecture is not specified


kunalspathak:
There is an issue in the script in getting the architecture. When I run this script on windows-arm64 machine with powershell (running on x64), the PROCESSOR_ARCHITECTURE value returned is AMD64 and it falsely installs x64 version. Is there a robust way to know what is the architecture of current machine?

YuliiaKovalova:
Hi @kunalspathak,

Thank you for providing the details.
I will investigate it and return to you with updates.

Issue #279, opened 2022-06-09
Improve logging when it's impossible to extract archive due to a file not allowing overwrite


YuliiaKovalova:
dotnet-install should give better logging when it can't extract due to a file not allowing overwrite for Windows environment
please read the [thread](https://teams.microsoft.com/l/message/19:afba3d1545dd45d7b79f34c1821f6055@thread.skype/1654640816284?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=4d73664c-9f2f-450d-82a5-c2f02756606d&parentMessageId=1654640816284&teamName=.NET%20Core%20Eng%20Services%20Partners&channelName=First%20Responders&createdTime=1654640816284)

PR #284, opened 2022-06-17
remove depricated test projects + small fixes

YuliiaKovalova:
- deprecated tests are removed
- logging for the case when the package can't be extracted is resolved [issue 279](https://github.com/dotnet/install-scripts/issues/279) 
- architecture definition for PS x64 on arm64 machine is fixed [issue 283](https://github.com/dotnet/install-scripts/issues/283)


src/dotnet-install.ps1
```diff
@@ -204,6 +204,15 @@ function Get-Machine-Architecture() {
         return $ENV:PROCESSOR_ARCHITEW6432
     }
 
+    # covers the case when PS x64 is run on ARM machine
+    if( (wmic os get osarchitecture)[2] -like "ARM*" ) {
+        if( [Environment]::Is64BitOperatingSystem )
```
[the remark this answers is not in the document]
YuliiaKovalova: the ternary operator was introduced in PS v7 https://toastit.dev/2019/09/25/ternary-operator-powershell-7/
I am afraid it can bring some problems with compatibility for some users

src/dotnet-install.ps1
```diff
@@ -805,6 +814,10 @@ function Extract-Dotnet-Package([string]$ZipPath, [string]$OutPath) {
             }
         }
     }
+    catch
+    {
+        Say-Error "Failed to extract package. Exception: $_"
+    }
```
bekir-ozturk, line 820: Not having this catch block was resulting in the exception bubble up, print the error and terminate the execution.
With this new change, we will print the exception, but continue execution and happily print "Installation succeeded" in the end.

We should probably rethrow.

src/dotnet-install.ps1
```diff
@@ -204,6 +204,15 @@ function Get-Machine-Architecture() {
         return $ENV:PROCESSOR_ARCHITEW6432
     }
 
+    # covers the case when PS x64 is run on ARM machine
+    if( (wmic os get osarchitecture)[2] -like "ARM*" ) {
```
bekir-ozturk, line 208: Apparently, [wmic is deprecated](https://docs.microsoft.com/en-us/windows/win32/wmisdk/wmic) starting Windows 10 21H1.
Arm windows machines are available starting Windows 11.
It may be better to use the [recommended approach](https://docs.microsoft.com/en-us/powershell/scripting/learn/ps101/07-working-with-wmi?view=powershell-7.2) to avoid technical dept.

bekir-ozturk (changes requested):
Being a big fan of deleting old files, looks nice! I have some suggestions on other changes. 
Please also consider splitting such work into multiple PRs in the future so that we can merge them independently.

src/dotnet-install.ps1

src/dotnet-install.ps1
```diff
@@ -204,6 +204,15 @@ function Get-Machine-Architecture() {
         return $ENV:PROCESSOR_ARCHITEW6432
     }
 
+    # covers the case when PS x64 is run on ARM machine
+    if( ((Get-WmiObject Win32_OperatingSystem).OSArchitecture) -like "ARM*") {
```
bekir-ozturk, line 208: `Get-WmiObject` also seems [outdated](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-wmiobject?view=powershell-5.1#description). The following should be the way if I'm not mistaken:
```suggestion
    if( ((Get-CimInstance -ClassName CIM_OperatingSystem).OSArchitecture) -like "ARM*") {
```

bekir-ozturk (approved):

PR #286, opened 2022-06-22
Fix check if version and quality are not specified

### LINKED ISSUE #285: dotnet-install.sh no longer working with both --version and --quality specified
### Description

Our bootstrapper script for PowerShell 7 has been specifying both the version and quality and has been working until a few days ago when it started returning an error.  If I modify our script to pass just the --version and leave out --quality it works.  This is a breaking change and the error message seems to indicate this is a bug and not intended behavior.  Note that the .ps1 doesn't have this problem.

### Reproduction Steps

./dotnet-install.sh -v 7.0.100-preview.5.22307.18 -q daily -skipnonversionedfiles

### Expected behavior

dotnet should install the version specified

### Actual behavior

dotnet_install: Error: Either Quality or Version option has to be specified. See https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script#options for details.

### Regression?

Yes, starting failing this week

### Known Workarounds

Remove --quality and only specify --version.  The PowerShell script dotnet-install.ps1 doesn't have this problem where we continue to specify both version and quality.

### Configuration

7.0.100-preview.5.22307.18

### Other information

_No response_

SteveL-MSFT:
Fix #285

This PR https://github.com/dotnet/install-scripts/pull/278 introduced a check and based on the error message it seems to be validating that at least --version or --quality is specified.  The current `if` statement doesn't do this correctly, it's comparing only if `--version latest` isn't specified and `-n` does the opposite of checking if `--quality` is empty.

Used https://linuxhint.com/testing_strings_bash_z_n as a resource on the difference of now `-n` and `-z` works for bash `if` statement for strings.  Where `-n` returns true if the string is NOT null, and `-z` returns true if the string IS null.

Note that are other usage of `-n` in the script and I didn't validate if those are correct.  I only tested this change locally that the way our CI uses dotnet-install.sh with both --version and --quality specified works.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -1190,7 +1190,7 @@ generate_akams_links() {
     local valid_aka_ms_link=true;
 
     normalized_version="$(to_lowercase "$version")"
-    if [[ "$normalized_version" != "latest" ]] && [ -n "$normalized_quality" ]; then
+    if [[ -z "$normalized_version" && -z "$normalized_quality" ]]; then
         say_err "Either Quality or Version option has to be specified. See https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script#options for details."
         return 1
     fi


SteveL-MSFT:
Also need the new version published to https://dotnet.microsoft.com/download/dotnet/scripts/v1/dotnet-install.sh to unblock PowerShell team CI

SteveL-MSFT:
@YuliiaKovalova please let me know what needs to be done to get this published to the download site.  Thanks.

YuliiaKovalova:
Hi @SteveL-MSFT ,

Sorry, I am on medical leave this week and can't deploy these changes. @bekir-ozturk and I had a short conversation yesterday regarding this issue and it seems like presence of the both params are invalid according to new rules.
I will bring you a decision on Monday.

I apologize for delay.

Issue #287, opened 2022-06-26
Unintended option checking with PR https://github.com/dotnet/install-scripts/pull/286


bekir-ozturk:
Specifying both `--quality` and `--version` options makes the caller's intent unclear since:
- If the user asks for build with a specific quality, the script determines the version for the given quality.
- If the user specifies a version, that means that s/he has already implicitly picked a quality since there is only ever one build with a given version number (with whatever quality it was built with).

Therefore, we have disallowed specifying both options within a single invocation with PR https://github.com/dotnet/website/pull/4088.
The commit https://github.com/dotnet/install-scripts/commit/fc0b11b5f0a15fc3a91d0c44fa36b64e5fb4afaa introduced with https://github.com/dotnet/install-scripts/pull/286 breaks the implementation in an unintended way.

I suggest that:
 - We revert the commit.
 - Clarify the error message since the current message can be interpreted as "at least one needs to be specified" rather than "at most one needs to be specified" (like it was done in [this comment](https://github.com/dotnet/install-scripts/issues/285#issuecomment-1162601593)).
 - Add some tests to make sure it works the intended way.

PR #288, opened 2022-06-27
revert 286 + extend test coverage

YuliiaKovalova:
revert 286 according to https://github.com/dotnet/install-scripts/issues/287, improve error message for case when quality and version are both specified, extend test coverage

Documentation is also updated 
[https://github.com/dotnet/docs/pull/30005](https://github.com/dotnet/docs/pull/30005)

### DIFF
--- a/tests/Install-Scripts.Test/GivenThatIWantToInstallDotnetFromAScript.cs
+++ b/tests/Install-Scripts.Test/GivenThatIWantToInstallDotnetFromAScript.cs
@@ -600,8 +600,7 @@ namespace Microsoft.DotNet.InstallationScript.Tests
         [InlineData("6.0.1", Quality.Signed)]
         public void WhenBothVersionAndQualityWereSpecified(string version, Quality quality)
         {
-            string feedCredentials = Guid.NewGuid().ToString();
-            var args = new[] { "-dryrun", "-version", version, "-quality", quality.ToString(), "-feedCredential", feedCredentials };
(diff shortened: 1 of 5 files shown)


tests/Install-Scripts.Test/GivenThatIWantToInstallDotnetFromAScript.cs
```diff
@@ -595,6 +595,24 @@ public void WhenFailingToInstallWithFeedCredentials(string? runtime, string chan
             commandResult.Should().NotHaveStdOutContainingIgnoreCase(feedCredential);
         }
 
+        [Theory]
+        [InlineData("7.0.1", Quality.Daily)]
+        [InlineData("6.0.1", Quality.Signed)]
+        public void WhenBothVersionAndQualityWereSpecified(string version, Quality quality)
```
bekir-ozturk, line 601: If we don't have already, I suggest adding cases where:
- Only quality is specified
- Only version is specified
- Neither is specified
and test that the error message is not displayed in all cases.
I suspect we already have these cases being used in the tests, but probably not all.

bekir-ozturk (commented):
Looks good!

tests/Install-Scripts.Test/GivenThatIWantToInstallDotnetFromAScript.cs
YuliiaKovalova: Done, thank you !

bekir-ozturk (approved):

PR #306, opened 2022-08-16
Support s390x

iii-i:
s390x support was introduced in .NET 6, and even though the
corresponding binaries are not distributed via
dotnetcli.azureedge.net, it is possible to get them from a custom feed.
When preparing a tarball for the source build, one could use this as
follows:

```
./build.sh -runtimesourcefeed /path/to/feed /p:DotNetRuntimeSourceFeed=/path/to/feed ...
```

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -306,6 +306,10 @@ get_machine_architecture() {
             echo "arm64"
             return 0
             ;;
+        s390x)
+            echo "s390x"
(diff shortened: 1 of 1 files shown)


YuliiaKovalova:
Hi @iii-i ,

Thank you for contributing to this project.
We will deploy updates soon.

YuliiaKovalova (approved):

Issue #308, opened 2022-08-19
dotnet-install.sh can't parse global.json containing CRLF


MihaZupan:
The `.sh` script will fail with if the `global.json` contains CRLF line endings.
```
Error: Unable to find the SDK:version node in `global.json`
```

Can the [json parsing function](https://github.com/dotnet/install-scripts/blob/6b1cc69894fa87684dbd65548f7679eed1a979e8/src/dotnet-install.sh#L517-L553) be updated to support this?

Currently this requires an ugly workaround of manually changing line endings, e.g.
```bash
sed 's/\r$//' global.json > global-lf.json
curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin --jsonfile global-lf.json
```

YuliiaKovalova:
Hi @MihaZupan,

Thank you for the ticket. We are working on resolving this issue.

YuliiaKovalova:
Hi @MihaZupan ,

It was fixed. I am planning to deploy it next week.

MihaZupan:
Thanks Yuliia

PR #309, opened 2022-08-23
fix parsing of line endings for bash script

YuliiaKovalova:
fix parsing of line endings for bash script - [#308](https://github.com/dotnet/install-scripts/issues/308)

### DIFF
--- a/tests/Install-Scripts.Test/Assets/InstallationScriptTestsWithWindowsLineEndings.json
+++ b/tests/Install-Scripts.Test/Assets/InstallationScriptTestsWithWindowsLineEndings.json
@@ -1,0 +1,6 @@
+{
+    "sdk": {
+        "version": "6.0.300",
+        "rollForward": "latestFeature"
+    }
+}
--- a/tests/Install-Scripts.Test/Assets/InstallationScriptTestsWithWindowsLineEndings.json
+++ b/tests/Install-Scripts.Test/Assets/InstallationScriptTestsWithWindowsLineEndings.json
@@ -1,6 +1,6 @@
 {
-    "sdk": {
-        "version": "6.0.300",
-        "rollForward": "latestFeature"
-    }
+  "sdk": {
+    "version": "1.0.0-beta.19463.3",
+    "rollForward": "latestFeature"
+  }
 }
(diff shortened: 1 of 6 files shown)


tests/Install-Scripts.Test/Assets/InstallationScriptTestsWithWindowsLineEndings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "sdk": {
+    "version": "1.0.0-beta.19463.3",
+    "rollForward": "latestFeature"
+  }
+}
```
bekir-ozturk, line 6: Depending on the configuration, `git` command may change the line endings. It is highly likely that this file will have `\n` line endings when checked-out to a CI machine on Linux.

I suggest that we add a `.gitattributes` file next to this json file and mark it as "always crlf" by:
```
InstallationScripttestsWithWindowsLineEndings.json text eol=crlf
```

tests/Install-Scripts.Test/GivenThatIWantToInstallDotnetFromAScript.cs
```diff
@@ -64,7 +64,8 @@ public class GivenThatIWantToInstallDotnetFromAScript : IDisposable
                 ("release/2.1.8xx", "2\\.1\\.8.*", Quality.None),
                 ("release/2.2.4xx", "2\\.2\\.4.*", Quality.None),
                 ("release/3.0.1xx", "3\\.0\\.1.*", Quality.None),
-               // ("release/3.1.4xx", "3\\.1\\.4.*", Quality.None),
+                // version is outdated
+                // ("release/3.1.4xx", "3\\.1\\.4.*", Quality.None),
```
bekir-ozturk, line 68: Can you elaborate on what exactly is outdated? 3.1.4xx is an LTS. 3.0.1xx isn't an LTS but it can be downloaded. Sounds like there is a different problem with 3.1.4 builds.

bekir-ozturk (changes requested):

tests/Install-Scripts.Test/GivenThatIWantToInstallDotnetFromAScript.cs
YuliiaKovalova: please find details here
https://github.com/dotnet/arcade/issues/10026

bekir-ozturk (approved):

PR #314, opened 2022-08-31
fix issue for osx arm64 when pre-6.0 is requested

YuliiaKovalova:
The fix was verified
https://github.com/dotnet/arcade/issues/10423#issuecomment-1233174764


src/dotnet-install.sh
```diff
@@ -353,6 +353,48 @@ get_normalized_architecture_from_architecture() {
...
+
+    local is_version_support_arm64="$(is_arm64_supported "$1")"
+    local is_channel_support_arm64="$(is_arm64_supported "$2")"
+    local architecture="$3";
+    local osname="$(get_current_os_name)"
+
+    if [ "$osname" == "osx" ] && [ "$architecture" == "arm64" ] && { [ "$is_version_support_arm64" = false ] || [ "$is_channel_support_arm64" = false ]; }; then
+        #check if rosetta is installed
+        if [ "$(/usr/bin/pgrep oahd >/dev/null 2>&1;echo $?)" -eq 0 ]; then 
+            say_verbose "Edit architecture for osx because SDK version is pre-6.0"
```
baronfel, line 371:
```suggestion
            say_verbose "Changing user architecture from '$architecture' to 'x64' because .NET SDKs prior to version 6.0 do not support arm64. 
```

src/dotnet-install.sh
```diff
@@ -353,6 +353,48 @@ get_normalized_architecture_from_architecture() {
...
+            echo "x64"
+            return 0;
+        else
+            say_err "Architecture \`$architecture\` is not supported for \`$version\` version. Please install Rosetta for resolving this issue."
```
baronfel, line 375:
```suggestion
            say_err "Architecture \`$architecture\` is not supported for .NET SDK version \`$version\`. Please install Rosetta to allow emulation of the \`$architecture\` .NET SDK on this platform"
```

baronfel (approved):
Some wording suggestions for the messages. Overall looks good, thank you for doing this!

PR #322, opened 2022-10-26
change channel value from current to sts

YuliiaKovalova:


src/dotnet-install.ps1
```diff
@@ -283,7 +283,7 @@ function Get-NormalizedChannel([string]$Channel) {
 
     switch ($Channel) {
         { $_ -eq "lts" } { return "LTS" }
-        { $_ -eq "current" } { return "current" }
+        { $_ -eq "sts" } { return "STS" }
```
bekir-ozturk, line 286: I don't think we should change the underlying channel name. What I understand from the earlier conversations around this was that this will be a purely cosmetic change. Otherwise, we would need to update aka.ms link generation and rename all the folders in the blob storage.

What we probably need is 
```suggestion
        { $_ -eq "sts" } { return "current" }
```

bekir-ozturk (changes requested):

src/dotnet-install.ps1
YuliiaKovalova: agree

tests/Install-Scripts.Test/AkaMsLinksTests.cs
```diff
@@ -318,13 +318,13 @@ public void LinkCanBeCreatedForGivenRuntime(string channel, string runtime, stri
...
+        [InlineData("STS", null, "daily", @"https://aka.ms/dotnet/current/dotnet-sdk-")]
         [InlineData("LTS", null, "signed", @"https://aka.ms/dotnet/LTS/dotnet-sdk-")]
-        [InlineData("Current", "dotnet", "validated", @"https://aka.ms/dotnet/current/dotnet-runtime-")]
+        [InlineData("STS", "dotnet", "validated", @"https://aka.ms/dotnet/current/dotnet-runtime-")]
         [InlineData("LTS", "dotnet", "preview", @"https://aka.ms/dotnet/LTS/dotnet-runtime-")]
-        [InlineData("Current", "aspnetcore", "daily", @"https://aka.ms/dotnet/current/aspnetcore-runtime-")]
+        [InlineData("STS", "aspnetcore", "daily", @"https://aka.ms/dotnet/current/aspnetcore-runtime-")]
         [InlineData("LTS", "aspnetcore", "signed", @"https://aka.ms/dotnet/LTS/aspnetcore-runtime-")]
-        [InlineData("Current", "windowsdesktop", "validated", @"https://aka.ms/dotnet/current/windowsdesktop-runtime-")]
+        [InlineData("STS", "windowsdesktop", "validated", @"https://aka.ms/dotnet/current/windowsdesktop-runtime-")]
```
bekir-ozturk, line 327: I believe we will need to continue supporting "current" channel since customers are still using it. If we remove tests related to "current" channel, we may break the related scenarios later and wouldn't notice it. I suggest that we keep the "Current" channel tests and add "STS" tests on the top.

bekir-ozturk: Maybe we can stop adding new tests with "current" starting with .NET 7.0 though.

src/dotnet-install.ps1
```diff
@@ -12,7 +12,7 @@
 .PARAMETER Channel
     Default: LTS
     Download from the Channel specified. Possible values:
-    - Current - most current release
+    - STS - standard term support SDK
```
bekir-ozturk, line 15: Some people are probably using "current" and they will not notice the change for a long time unless they check the documentation.

Should we add a warning to say "'Current' channel option is being deprecated. Use 'sts' instead."?

YuliiaKovalova: done

src/dotnet-install.ps1
```diff
@@ -277,13 +278,17 @@ function Get-NormalizedChannel([string]$Channel) {
         return ""
     }
 
+    if ($Channel.Contains("Current")) {
+        Say-Error 'Value "Current" was depricated for -Channel option. Please use "STS" instead.'
```
bekir-ozturk, line 282: Outputting to error stream will fail a lot of build pipelines. How about just a warning?
Typo: deprecated.

bekir-ozturk (approved):

PR #340, opened 2022-12-07
Added time measurements for several actions

MichalPavlik:
We want to know, how much time it takes for some actions.


baronfel:
After this change goes out, I would like to coordinate with the [actions/setup-dotnet](https://github.com/actions/setup-dotnet) repo to hook up verbose logging with the GitHub Actions [step debug logging](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging#enabling-step-debug-logging) capability. Ideally, when a user opts into step-level debugging, we should enable verbose logging in the install scripts automatically.

Issue #345, opened 2023-01-24
eng/common is not in sync with dotnet/arcade


Sapana-Khemkar:
eng/common folder code is not sync latest[ dotnet/arcade](https://github.com/dotnet/arcade) 
Latest arcade folder has support for ppc64le architecture which is missing in this repo

PR #346, opened 2023-01-24
add support for ppc64le in install-script

Sapana-Khemkar:
.NET 7 supports ppc64le architecture. Microsoft does not generate and uploads binaries at dotnetcli.azureedge.net as it is community supported architecture. But it is possible to get them from a custom feed.
When preparing a tarball for the source build, one could use this as follows:

`./build.sh -runtimesourcefeed /path/to/feed /p:DotNetRuntimeSourceFeed=/path/to/feed ...`

Note: eng/common folder is yet not in sync with latest arcade. Issue https://github.com/dotnet/install-scripts/issues/345 has been opened for the same.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -310,6 +310,10 @@ get_machine_architecture() {
             echo "s390x"
             return 0
             ;;
+        ppc64le)
+            echo "ppc64le"
+            return 0
+            ;;
         esac
     fi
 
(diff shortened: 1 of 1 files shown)


MichalPavlik:
cc: @baronfel

baronfel:
If we have an established pattern for arcade to enable platforms that MIcrosoft itself doesn't support as a first-class target, and folks are ok with getting errors when they request an architecture that isn't available on the chosen feed (which is the case today), then this seems good to take. I'm doing a bit of checking up on that before I give a proper thumbs-up on this.

YuliiaKovalova (approved):

PR #356, opened 2023-03-07
Reduce verbosity of the script by default

### LINKED ISSUE #351: `install-dotnet.ps1` is too verbose without way to quiet it
My global.json looks like this
```json
{
  "sdk": {
    "version": "8.0.100-preview.2.23115.11"
  },
  "tools": {
    "dotnet": "8.0.100-preview.2.23115.11",
    "runtimes": {
      "dotnet/x64": [
        "3.1.32",
        "6.0.13",
        "7.0.2"
      ],
      "aspnetcore/x64": [
        "3.1.32",
        "6.0.13",
        "7.0.2"
      ]
    }
  }
}
```

And when I restore the toolset (essentially invoking `install-dotnet.ps1` under covers) the output is too verbose:
image

And there's no way to suppress this notification:
https://github.com/dotnet/install-scripts/blob/590932815dc4104e9bec3fa517e247b07b7792ac/src/dotnet-install.ps1#L1113-L1116
https://github.com/dotnet/install-scripts/blob/590932815dc4104e9bec3fa517e247b07b7792ac/src/dotnet-install.ps1#L125-L133

I'd like to be able to disable this notification. Perhaps, the only information could be valuable (~ironically, the most important part of it's missing in `Attempting to install dotnet from public location.`~ this is the Arcade SDK's doing) is something like this:
```
dotnet-install: .NET Core Runtime with version '3.1.32' is already installed.
dotnet-install: .NET Core Runtime with version '6.0.13' is already installed.
dotnet-install: .NET Core Runtime with version '7.0.2' is already installed.
dotnet-install: ASP.NET Core Runtime with version '3.1.32' is already installed.
dotnet-install: ASP.NET Core Runtime with version '6.0.13' is already installed.
dotnet-install: ASP.NET Core Runtime with version '7.0.2' is already installed.
```

I'd venture to say that emitting this text by default we're wasting millions of hours of compute. This notification should be in the docs and shown when `-help` is invoked.

RussKie:
Add the script's purpose to the help text

Resolves #351

# Before
image

# After
image
image
image

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -9,6 +9,12 @@
 .DESCRIPTION
     Installs dotnet cli. If dotnet installation already exists in the given directory
     it will update it only if the requested version differs from the one already installed.
+
+    Note that the intended use of this script is for Continuous Integration (CI) scenarios, where:
+    - The SDK needs to be installed without user interaction and without admin rights.
+    - The SDK installation doesn't need to persist across multiple CI runs.
+    To set up a development environment or to run apps, use installers rather than this script. Visit https://dotnet.microsoft.com/download to get the installer.
+
 .PARAMETER Channel
     Default: LTS
     Download from the Channel specified. Possible values:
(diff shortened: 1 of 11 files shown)


donJoseLuis (approved):

MichalPavlik:
Hey @RussKie, thanks for the PR. Could you please also modify tests. It seems some of them are expecting to have removed text in stdout.

RussKie:
Tests amended, new added. Should be good to go now.
Thanks.

Issue #360, opened 2023-04-21
Possibility to output installed version from the scripts


IvanZosimov:
**Description:**
Is it possible after the run of `install-script` somehow output (e.g. set several environment variables) the resolved and installed version of the .NET SDK and runtime? It'd be very suitable for CI purposes.

For instance, after execution I'd like to be able to do this:
```
echo $Env:INSTALLED_DOTNET_SDK
echo $Env:INSTALLED_DOTNET_RUNTIME
```
If this feature is approved, I'd love to prepare pull request.

AR-May:
Team Triage: This feature sounds reasonable. Different 'products' can be installed by the scripts, so the format of the generated environment variables should be something like: `DOTNET_INSTALL_SCRIPTS_<PRODUCT>_LAST_INSTALLED_VERSION=<version>`.  Unfortunately, the project now is in the maintenance mode, and we do not have capacity for implementing this feature. However, we would be glad to accept such a feature if you would like to prepare a pull request.

PR #361, opened 2023-04-27
Add possibility to output an env. variables with installed by the script .NET product version

IvanZosimov:
**Description**:
In the scope of this PR, the function `Set-Informational-Env-Variable() / set_informational_env_variable()` was added both to the `powershell` and `bash` scripts. 

**Notes:**
While the Powershell script is working perfectly, the Bash script isn't as good. Setting env. variables in the parent shell is possible only using sourcing (source or dot command) but the Bash script contains a lot of exits. As soon as the script is executed in source mode and meets the exit() command the script closes the shell. If you could give me a hint how to workaround this bash "feature", it would be really nice!

**Related issue**:
https://github.com/dotnet/install-scripts/issues/360

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -1116,6 +1116,19 @@ function Prepare-Install-Directory {
     }
 }
 
+function Set-Informational-Env-Variable([string]$assetName, [string]$version ) {
+    $envVariablePostfix = $null
+    if ($assetName.ToLower().Contains("runtime")) {
+        $envVariablePostfix = "RUNTIME"
+    }
+    else {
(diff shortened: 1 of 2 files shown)


IvanZosimov:
> 

@dotnet-policy-service agree

AR-May (changes requested):

YuliiaKovalova:
Hi @IvanZosimov,

Do you plan to finalize this PR?

IvanZosimov:
Hi, @YuliiaKovalova 👋 Unfortunately, I don't. I found yet another way how to solve the issue. I'm closing this PR as well as the issue.

PR #379, opened 2023-08-14
change diskspace check due to PS issues

### LINKED ISSUE #374: C# plugin fails to load because 'dotnet-install.ps1'  fails with diskspace error
The issue text generated by visual studio is an empty file. Digging into the issue, it looks like `dotnet-install.ps1` is trying to determine how much disk space is available and throws an error if no disk space is available even though I have ~300GB free.

## Issue:
`Get-PSDrive`  is called in script `dotnet-install.ps1` function `Prepare-Install-Directory`. `Get-PSDrive` appears to return null for Free and 0 for Used in certain scenarios even when there is plenty of space on the drive.

```ps1
PS C:\Users\MyUser> Get-PSDrive -N C | select Free,Used
Free Used
 ---- ----
        0 
```

## Requested fix:
In `dotnet-install.ps1` function `Prepare-Install-Directory`, before the check to throw the error. Could yall check to see if Free and Used are null or zero and just display a warning instead? If the disk has 0 used data and null free data, that should be a signal that the `Get-PSDrive` call is not functioning correctly and can't be trusted.



## Work around:
Comment out the line that throws the exception and mark the file as read only. 
 * Open file: `C:\Users\MyUser\.vscode\extensions\ms-dotnettools.vscode-dotnet-runtime-1.6.0\dist\install scripts\dotnet-install.ps1`
 * Search for the following line and comment it out: ``throw "There is not enough disk space on drive ${installDrive}:"``
    * Comment out with a # in powershell.
    * dotnet-install.ps1:1115 in function "Prepare-Install-Directory" at the time of writing this ticket.
 * Right click file -> Properties -> click checkbox "Read Only"


## Snippets
The issue is in the function below:
dotnet-install.ps1:1102
```ps1
function Prepare-Install-Directory {
    New-Item -ItemType Directory -Force -Path $InstallRoot | Out-Null

    $installDrive = $((Get-Item $InstallRoot -Force).PSDrive.Name);
    $diskInfo = $null
    try{
    }
    catch{
        Say-Warning "Failed to check the disk space. Installation will continue, but it may fail if you do not have enough disk space."
    }
    # HERE: Requested fix here, add a check to see if Used and Free are both null/zero. If they are skip the below if block that can throw an error and just display a warning.
    # HERE: if (diskInfo.Free == 0 && diskInfo.Used == 0) {Say-Warning "..." and return}
    
    # HERE: Error preventing usage being thrown here.
    if ( ($null -ne $diskInfo) -and ($diskInfo.Free / 1MB -le 100)) {
        throw "There is not enough disk space on drive ${installDrive}:" # HERE: Comment this line out as a work around.
    }
}
```


Full `Get-PSDrive` output:
```
PS C:\Users\MyUser> Get-PSDrive -N C | select *


Used            : 0
Free            :
CurrentLocation : Users\MyUser
Name            : C
Provider        : Microsoft.PowerShell.Core\FileSystem
Root            : C:\
Description     : Windows
MaximumSize     :
Credential      : System.Management.Automation.PSCredential
DisplayRoot     :
```

My C drive, which vscode is installed on, has plenty of free space (around 300GB).
```
PS C:\Users\MyUser> [System.IO.DriveInfo]::GetDrives() | Format-Table

Name DriveType DriveFormat IsReady AvailableFreeSpace TotalFreeSpace     TotalSize RootDirectory VolumeLabel
---- --------- ----------- ------- ------------------ --------------     --------- ------------- -----------
C:\      Fixed NTFS           True       355307118592   355307118592 1023235059712 C:\           Windows
```
OS Version
```
PS C:\Users\MyUser> [System.Environment]::OSVersion.Version

Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      19044  0                                                                                                                                                                                                                             
```

YuliiaKovalova:
## Fixes 
#374

## Solution
For older PS versions it is a known issue: 
[build.cmd error dotnet-install: There is not enough disk space on drive C · Issue #35301 · dotnet/runtime (github.com)](https://github.com/dotnet/runtime/issues/35301)

[Get-PSDrive C returns blank used/free disk space · Issue #12442 · PowerShell/PowerShell (github.com)](https://github.com/PowerShell/PowerShell/issues/12442)

This fix loosens the check for PS with versions < 7.

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -1100,17 +1100,24 @@ function Resolve-AssetName-And-RelativePath([string] $Runtime) {
 }
 
 function Prepare-Install-Directory {
+    $diskSpaceWarning = "Failed to check the disk space. Installation will continue, but it may fail if you do not have enough disk space.";
+    # this check is relevant for PS version >= 7, the result can be irrelevant for older versions. See https://github.com/PowerShell/PowerShell/issues/12442.
+    if ($PSVersionTable.PSVersion.Major -lt 7) {
+        Say-Warning $diskSpaceWarning
+        return $null
+    }
+
     New-Item -ItemType Directory -Force -Path $InstallRoot | Out-Null
 
     $installDrive = $((Get-Item $InstallRoot -Force).PSDrive.Name);
     $diskInfo = $null
-    try{
+    try {
         $diskInfo = Get-PSDrive -Name $installDrive
     }
-    catch{
-        Say-Warning "Failed to check the disk space. Installation will continue, but it may fail if you do not have enough disk space."
+    catch {
+        Say-Warning $diskSpaceWarning
     }
-    
+
     if ( ($null -ne $diskInfo) -and ($diskInfo.Free / 1MB -le 100)) {
         throw "There is not enough disk space on drive ${installDrive}:"
     }
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -1101,7 +1101,7 @@ function Resolve-AssetName-And-RelativePath([string] $Runtime) {
 
 function Prepare-Install-Directory {
     $diskSpaceWarning = "Failed to check the disk space. Installation will continue, but it may fail if you do not have enough disk space.";
-    # this check is relevant for PS version >= 7, the result can be irrelevant for older versions. See https://github.com/PowerShell/PowerShell/issues/12442.
+
     if ($PSVersionTable.PSVersion.Major -lt 7) {
         Say-Warning $diskSpaceWarning
         return $null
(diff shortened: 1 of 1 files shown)


src/dotnet-install.ps1
```diff
@@ -1100,17 +1100,24 @@ function Resolve-AssetName-And-RelativePath([string] $Runtime) {
 }
 
 function Prepare-Install-Directory {
+    $diskSpaceWarning = "Failed to check the disk space. Installation will continue, but it may fail if you do not have enough disk space.";
+    # this check is relevant for PS version >= 7, the result can be irrelevant for older versions. See https://github.com/PowerShell/PowerShell/issues/12442.
+    if ($PSVersionTable.PSVersion.Major -lt 7) {
+        Say-Warning $diskSpaceWarning
```
baronfel, line 1106: I wasn't sure about using `Warning` here because the behavior depends on the callers' warning preference, but the SDK Acquisition extension can control this on its end, so I think this is acceptable here :+1:

baronfel (approved):

PR #386, opened 2023-08-29
Add logging for remote/local package sizes

### LINKED ISSUE #382: Add HTTP head log entry
getting the targeted file size can offer insight into the corrupt bits observed in vscode regional cases.

exit criteria
1. NEW STEP: resolve targeted URI to download
2. NEW STEP: make HTTP head request for URI
3. NEW STEP: log the content-length header (file size in bits)
4. download the target (as it currently works)
5. NEW STEP: log the downloaded file size in bits
6. continue with installation

Example of http head for resolved URI:

>curl --head https://dotnetcli.azureedge.net/dotnet/Sdk/8.0.100-preview.7.23376.3/dotnet-sdk-8.0.100-preview.7.23376.3-win-x64.zip
HTTP/1.1 200 OK
Accept-Ranges: bytes
Age: 442119
Content-Type: application/octet-stream
Date: Fri, 25 Aug 2023 10:54:09 GMT
Etag: 0x8DB9808765EAB0C
Last-Modified: Tue, 08 Aug 2023 12:10:31 GMT
Server: ECAcc (frc/4CA2)
X-Cache: HIT
x-ms-blob-type: BlockBlob
x-ms-lease-status: unlocked
x-ms-request-id: 0b2a2b71-801e-0017-0e3d-d3f80a000000
x-ms-version: 2009-09-19
Content-Length: 281161152

YuliiaKovalova:
Fixes: #382

Logging for file sizes is added for tracking corruption cases.

Checked manually on Windows/Mac machines

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -176,6 +176,16 @@ function Measure-Action($name, $block) {
     Say-Verbose "⏱ Action '$name' took $totalSeconds seconds"
 }
 
+function Get-File-Size($zipUri) {
+    $response = Invoke-WebRequest -Uri $zipUri -Method Head
+    $fileSize = [long]$response.Headers["Content-Length"]
+    $fileSizeBits = $fileSize * 8
+
+    Say "Initial file $zipUri size is $fileSizeBits bits."
+
+    return $fileSizeBits
+}
+
 function Say-Invocation($Invocation) {
     $command = $Invocation.MyCommand;
     $args = (($Invocation.BoundParameters.Keys | foreach { "-$_ `"$($Invocation.BoundParameters[$_])`"" }) -join " ")
@@ -863,16 +873,30 @@ function DownloadFile($Source, [string]$OutPath) {
 
     $Stream = $null
 
+    $remoteFileSize = Get-File-Size -zipUri $Source
+    
     try {
         $Response = GetHTTPResponse -Uri $Source
         $Stream = $Response.Content.ReadAsStreamAsync().Result
         $File = [System.IO.File]::Create($OutPath)
         $Stream.CopyTo($File)
         $File.Close()
+
+        # Calculate the downloaded file size in bits
+        $fileSizeBytes = [long](Get-Item $OutPath).Length
+        $fileSizeBits = $fileSizeBytes * 8
+
+        # Log the downloaded file size in bits
+        Say "Downloaded file $Source size is $fileSizeBits bits."
     }
     finally {
         if ($null -ne $Stream) {
             $Stream.Dispose()
+        }
+
+        if ($remoteFileSize -ne $fileSizeBits)
+        {
+            Say "The remote and local file sizes are not equal for $Source. The downloaded package may be corrupted."
         }
     }
 }
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -181,10 +181,9 @@ function Get-Remote-File-Size($zipUri) {
         $response = Invoke-WebRequest -Uri $zipUri -Method Head
         $fileSize = $response.Headers["Content-Length"]
         if ((![string]::IsNullOrEmpty($fileSize))) {
-            $fileSizeBits = [long]$fileSize * 8
-            Say "Initial file $zipUri size is $fileSizeBits bits."
+            Say "Remote file $zipUri size is $fileSize bytes."
         
-            return $fileSizeBits
+            return $fileSize
         }
      
     }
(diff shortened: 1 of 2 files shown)


src/dotnet-install.ps1
```diff
@@ -176,6 +176,16 @@ function Measure-Action($name, $block) {
     Say-Verbose "⏱ Action '$name' took $totalSeconds seconds"
 }
 
+function Get-File-Size($zipUri) {
+    $response = Invoke-WebRequest -Uri $zipUri -Method Head
+    $fileSize = [long]$response.Headers["Content-Length"]
+    $fileSizeBits = $fileSize * 8
```
ladipro, line 182: Why are the sizes converted to bits?

ladipro (commented):
I've left comments inline but I also have a concern with the feature itself. The purpose of the `Content-Length` header is to communicate the size of the data being transferred _after_ it's been encoded. What this means is that it can be perfectly legal for `Content-Length` to be different from the downloaded file if the request uses e.g. compression. It is also perfectly legal for this header to not be sent by the server.

Can you please implement it more defensively? Specifically, it should be fine for the HEAD request to fail (it's not guaranteed to be supported), and it should be fine for the `Content-Length` response header to not be present.

src/dotnet-install.ps1
YuliiaKovalova: @donJoseLuis did you have any specific reason for requesting this?

src/dotnet-install.ps1
```diff
@@ -863,17 +873,31 @@ function DownloadFile($Source, [string]$OutPath) {
 
     $Stream = $null
 
+    $remoteFileSize = Get-File-Size -zipUri $Source
```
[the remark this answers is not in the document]
donJoseLuis: +1  (new logging should be harmless to the execution, as Ladi mentioned)

src/dotnet-install.ps1
```diff
@@ -863,17 +882,30 @@ function DownloadFile($Source, [string]$OutPath) {
...
+        # Log the downloaded file size in bits
+        Say "Downloaded file $Source size is $fileSizeBits bits."
     }
     finally {
         if ($null -ne $Stream) {
             $Stream.Dispose()
         }
+
+        if ((![string]::IsNullOrEmpty($remoteFileSize)) -and $remoteFileSize -ne $fileSizeBits) {
+            Say "The remote and local file sizes are not equal."
```
donJoseLuis, line 907: minor: please consider outputting the sizes: source and actual.

donJoseLuis (changes requested):

src/dotnet-install.ps1

donJoseLuis (approved):

src/dotnet-install.ps1
```diff
@@ -176,6 +176,24 @@ function Measure-Action($name, $block) {
...
+function Get-Remote-File-Size($zipUri) {
+    try {
+        $response = Invoke-WebRequest -Uri $zipUri -Method Head
+        $fileSize = $response.Headers["Content-Length"]
+        if ((![string]::IsNullOrEmpty($fileSize))) {
+            Say "Remote file $zipUri size is $fileSize bytes."
+        
+            return $fileSize
+        }
+
```
ladipro, line 188: super-nit: Extra empty line.

src/dotnet-install.ps1
```diff
@@ -176,6 +176,24 @@ function Measure-Action($name, $block) {
...
+        $fileSize = $response.Headers["Content-Length"]
+        if ((![string]::IsNullOrEmpty($fileSize))) {
+            Say "Remote file $zipUri size is $fileSize bytes."
+        
+            return $fileSize
+        }
+     
+    }
+    catch {
+        Say-Verbose "Content-Length header was not extacted for $zipUri."
```
ladipro, line 191: Typo: extacted -> extracted ?

src/dotnet-install.ps1
```diff
@@ -863,17 +881,29 @@ function DownloadFile($Source, [string]$OutPath) {
...
+
+        $fileSize = [long](Get-Item $OutPath).Length
+        Say "Downloaded file $Source size is $fileSize bytes."
     }
     finally {
         if ($null -ne $Stream) {
             $Stream.Dispose()
         }
+
+        if ((![string]::IsNullOrEmpty($remoteFileSize)) -and $remoteFileSize -ne $fileSize) {
```
ladipro, line 901: Add `-and ![string]::IsNullOrEmpty($fileSize)` so we don't print the message if anything in the try block failed. Or maybe better, move the new code inside the try block? There's no point executing this code if the download has failed.

src/dotnet-install.sh
```diff
@@ -546,6 +546,36 @@ is_dotnet_package_installed() {
...
+        file_size="$(stat -f '%z' "$downloaded_file")"
+    fi  
+    
+    if [ -n "$file_size" ]; then
+        say "Downloaded file size is $file_size bytes."
+
+        if [ -n "$remote_file_size" ] && [ "$file_size" != "$remote_file_size" ]; then
+            say "The remote and local file sizes are not equal. Remote file size is $remote_file_size bytes and local size is $file_size bytes. The local package may be corrupted."
+        else
+            say "The remote and local file sizes are equal."
```
ladipro, line 572: The way the conditions are written now, we print this message if we were not able to get the remote size. This looks unexpected.

src/dotnet-install.sh
```diff
@@ -914,14 +944,39 @@ copy_files_or_dirs_from_list() {
...
+    else
+        say "Neither curl nor wget is available on this system."
+        return
+    fi
+
+    if [ -n "$file_size" ]; then
+        say "Remote file $zip_uri size is $file_size bytes."
+        echo "$file_size"
+    else
+        say_verbose "Content-Length header was not extacted for $zip_uri."
```
ladipro, line 965: Typo: extacted.

ladipro (changes requested):
Thank you for the updates. I've added a few more comments.

ladipro (approved):
👍

Issue #391, opened 2023-09-08
Incorrect logging about size of file


BrennanConroy:
dotnet-install.sh incorrectly logs:
> The remote and local file sizes are not equal. Remote file size is 31367277
 bytes and local size is 31367277 bytes. The local package may be corrupted.

Caused by https://github.com/dotnet/install-scripts/pull/386/files
@YuliiaKovalova

YuliiaKovalova:
hi @BrennanConroy,

thank you for reporting the issue, I am investigating it.

YuliiaKovalova:
The fix was merged & deployed.

PR #393, opened 2023-09-14
fix remote vs local file comparison in bash script

YuliiaKovalova:
## Problem 

https://github.com/dotnet/install-scripts/issues/391

## Solution
`file_size` variable is an integer, and `remote_file_size` is a string. For the correct comparison, there is a need to cast `remote_file_size`to integer.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -562,8 +562,8 @@ validate_remote_local_file_sizes()
     eval $invocation
 
     local downloaded_file="$1"
-    local remote_file_size="$2"
-
+    local remote_file_size=$(echo "$2" | awk '{ num = $1 + 0; print num }')
+    
     local file_size=''
     if [[ "$OSTYPE" == "linux-gnu"* ]]; then
         file_size="$(stat -c '%s' "$downloaded_file")"
@@ -575,7 +575,7 @@ validate_remote_local_file_sizes()
         say "Downloaded file size is $file_size bytes."
 
         if [ -n "$remote_file_size" ] && [ -n "$file_size" ]; then
-            if [ "$file_size" -ne "$(string_to_int "$remote_file_size")" ]; then
+            if [ "$file_size" -ne "$remote_file_size" ]; then
                 say "The remote and local file sizes are not equal. The remote file size is $remote_file_size bytes and the local size is $file_size bytes. The local package may be corrupted."
             else
                 say "The remote and local file sizes are equal."
@@ -585,11 +585,6 @@ validate_remote_local_file_sizes()
     else
         say "Either downloaded or local package size can not be measured. One of them may be corrupted."      
     fi 
-}
-
-# Function to convert a string to an integer
-function string_to_int {
-    echo "$1" | tr -d '\n' | sed 's/[^0-9]*//g'
 }
 
 # args:
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -563,7 +563,8 @@ validate_remote_local_file_sizes()
 
     local downloaded_file="$1"
     local remote_file_size="$2"
-    local file_size=0
+    local file_size=''
+
     if [[ "$OSTYPE" == "linux-gnu"* ]]; then
         file_size="$(stat -c '%s' "$downloaded_file")"
     elif [[ "$OSTYPE" == "darwin"* ]]; then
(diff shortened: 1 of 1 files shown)


src/dotnet-install.sh
```diff
@@ -587,6 +587,11 @@ validate_remote_local_file_sizes()
     fi 
 }
 
+# Function to convert a string to an integer
+function string_to_int {
+    echo "$1" | tr -d '\n' | sed 's/[^0-9]*//g'
```
ladipro, line 592: What does the problematic `remote_file_size` look like? If it's the trailing newline that's causing the issue, I believe it's fixable when the variable is assigned to.

Would this help in `get_remote_file_size`?
```bash
... | awk '{ORS=""; print $2}'
```

ladipro: The `ORS=""` part specifically.

YuliiaKovalova: > The `ORS=""` part specifically.

`remote_file_size` looks like "123", when "file_size" is 123. 
Unfortunately, the suggested` ORS="" `didn't help. on `[ "$file_size" -ne "$remote_file_size" ]` it causes "integer expression expected", on `[ "$file_size" != "$remote_file_size" ]` returns invalid comparison

ladipro: Ah, so if we need to get rid of the quotes, then:

```bash
... | awk '{gsub("\"","",$2); print $2}'
```

YuliiaKovalova: I used a bit different thing, please take a look

ladipro: Looks ok but I still think it should happen as part of the original assignment to remote_file_size in:

https://github.com/dotnet/install-scripts/blob/cab6fcd7d451b3284e45b6eff7f1fd8d324ece17/src/dotnet-install.sh#L964,L966

- It would be consistent with the variable holding the local size (both are integers from the get go).
- As a bonus it will run with one less process created (we already use `awk` there).

src/dotnet-install.sh
```diff
@@ -961,9 +961,9 @@ get_remote_file_size() {
     local zip_uri="$1"
 
     if machine_has "curl"; then
-        file_size=$(curl -sI  "$zip_uri" | grep -i content-length | awk '{print $2}')
+        file_size=$(curl -sI  "$zip_uri" | grep -i content-length | awk '{ num = $2 + 0; print num }')
     elif machine_has "wget"; then
-        file_size=$(wget --server-response -O /dev/null "$zip_uri" 2>&1 | grep -i '^Content-Length:' | awk '{print $2}')
+        file_size=$(wget --server-response -O /dev/null "$zip_uri" 2>&1 | grep -i '^Content-Length:' | awk '{ num = $2 + 0; print num }')
```
ladipro, line 966: I am testing this locally and the `wget` branch seems to be broken. Can you please confirm that this command works for you? My `wget` prints headers indented so the `grep` regex won't match it.

```
~$ wget --server-response -O /dev/null "http:some.random.url"
--2023-09-15 16:44:40--  http://some.random.url
Connecting to 192.168.1.9:80... connected.
HTTP request sent, awaiting response...
  HTTP/1.1 200 OK
  Date: Fri, 15 Sep 2023 14:44:40 GMT
  Server: Apache/2.4.41 (Ubuntu)
  Last-Modified: Fri, 15 Sep 2023 09:36:51 GMT
  ETag: "2e99060-60562883e92c0"
  Accept-Ranges: bytes
  Content-Length: 48861280
  Keep-Alive: timeout=5, max=100
  Connection: Keep-Alive
  Content-Type: application/vnd.android.package-archive
Length: 48861280 (47M) [application/vnd.android.package-archive]
Saving to: '/dev/null'
```

src/dotnet-install.sh
ladipro: Apologies for scrutinizing but it's important to understand what the exact issue is we're fixing. Apparently it's not newline and it's not quotes:

```
~$ curl -sI http://some.random.url | grep -i content-length | awk '{print $2}' | xxd
00000000: 3438 3836 3132 3830 0d0a                 48861280..
~$ curl -sI http://some.random.url | grep -i content-length | awk '{num = $2 + 0; print num}' | xxd
00000000: 3438 3836 3132 3830 0a
```

It's the CR (0d) character that throws the shell off because HTTP uses the CRLF "Windows-style" new-line sequence.

src/dotnet-install.sh
YuliiaKovalova: For wget we don't need "^", it's a good catch!
image

YuliiaKovalova: curl works as expected
image

ladipro (approved):

PR #394, opened 2023-09-15
add linter and fix listed issues 

YuliiaKovalova:


YuliiaKovalova:
all issues for PS are addressed, but left issues in the bash script require refactoring of the logic.

Issue #396, opened 2023-09-16
`stat` call fails on macOS when the GNU stat is used instead of BSD stat


daleeidd:
# Describe the bug

Get the following:
>  Error!
> .NET Acquisition Failed: Installation failed: stat: cannot read file system information for '%z': No such file or directory

## **Please attach log file(s) here if you have any:**

[DotNetAcquisition-ms-dotnettools.vscode-dotnet-runtime-1694845816479.txt](https://github.com/dotnet/vscode-dotnet-runtime/files/12640022/DotNetAcquisition-ms-dotnettools.vscode-dotnet-runtime-1694845816479.txt)

Output error:
> visualstudiotoolsforunity.vstuc requested to download the .NET Runtime.
> Downloading .NET version(s) 7.0.11 ......... Error!
> Failed to download .NET 7.0.11:
> stat: cannot read file system information for '%z': No such file or directory
>
> Error!
> .NET Acquisition Failed: Installation failed: stat: cannot read file system information for '%z': No such file or directory

## Other Info:

**Reproduction Steps**

Open any C# project.

**Further technical details**

I have the GNU Tools first in my path so using their version of stat.

Stat:
> stat (GNU coreutils) 9.3
> Copyright (C) 2023 Free Software Foundation, Inc.
> License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
> This is free software: you are free to change and redistribute it.
> There is NO WARRANTY, to the extent permitted by law.
> 
> Written by Michael Meskes.

VSCode
> Version: 1.81.1 (Universal)
> Commit: 6c3e3dba23e8fadc360aed75ce363ba185c49794
> Date: 2023-08-09T22:20:33.924Z (1 mo ago)
> Electron: 22.3.18
> ElectronBuildId: 22689846
> Chromium: 108.0.5359.215
> Node.js: 16.17.1
> V8: 10.8.168.25-electron.0
> OS: Darwin arm64 22.6.0

Extension:
> 1.7.3

baronfel:
Folks, we need to see if there's a way to detect if `stat` is GNU or BSD version - the arguments will change depending on which one we have.

craigds:
could just hardcode the path to `/usr/bin/stat` if you always want the BSD one

baronfel:
Closing as fixed in #397. The scripts served from the website will be updated soon.

PR #397, opened 2023-09-22
use stat from /usr/bin/stat explicitly

YuliiaKovalova:
## Fixes:
https://github.com/dotnet/install-scripts/issues/396

## Solution:
GNU stat can be specified in PATH before BSD one.
stat from /usr/bin/stat is needed for our purposes.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -567,9 +567,9 @@ validate_remote_local_file_sizes()
 
     if [[ "$OSTYPE" == "linux-gnu"* ]]; then
         # hardcode in order to avoid conflicts with GNU stat
-        file_size="$(/usr/bin/stat -c '%s' "$downloaded_file")"
+        file_size="$(stat -c '%s' "$downloaded_file")"
     elif [[ "$OSTYPE" == "darwin"* ]]; then
-        file_size="$(stat -f '%z' "$downloaded_file")"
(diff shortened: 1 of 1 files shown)


src/dotnet-install.sh
```diff
@@ -566,7 +566,8 @@ validate_remote_local_file_sizes()
     local file_size=''
 
     if [[ "$OSTYPE" == "linux-gnu"* ]]; then
-        file_size="$(stat -c '%s' "$downloaded_file")"
+        # hardcode in order to avoid conflicts with GNU stat
```
baronfel, line 569: This is the wrong branch - we want to use the BSD stat on macOS, not on Linux, since homebrew on macOS can install the GNU versions.

baronfel (changes requested):

src/dotnet-install.sh
baronfel: If we have a Mac for testing, you can verify the bad behavior by installing https://formulae.brew.sh/formula/coreutils and running our scripts.

YuliiaKovalova: fixed.

baronfel (approved):
Thanks! As we spoke about it would be good to be able to test on mac in an environment where we have the GNU coreutils package installed, but I'm not sure how much ability we have to change the azdo build environment.

Issue #402, opened 2023-10-10
Allow skipping the productVersion.txt check


baronfel:
In https://github.com/dotnet/install-scripts/issues/85 a change was made that allowed for mapping from a user-facing version (e.g. 5.0.100) to a 'final' product version (e.g. 5.0.100-rtm.20471.1) to allow for easier use of the scripts. This would allow users to use the 'friendly' versions of SDKs/Runtimes while still ensuring that they got the most up to date final construction. This check is done as part of `Get-Product-Version`/`get_specific_product_version_url` and happens all of the time. The current logic is something like:

* Check feed for a productVersion.txt file for the provided logical version - if present use that version
* If not present, check the generated download link and parse a version from it - if present use that version
* otherwise use the user-passed-in version

This works great for Arcade's use case, but it causes problems when used with another common use case: setting up a parallel set of proxy URLs to use as a custom feed.  

To address download failures with the current feeds used by the Install Scripts, the team from https://github.com/dotnet/vscode-dotnet-runtime wanted to set up a series of https://aka.ms links that would emulate the URI structure expected by these scripts, using destination URLs that point to the VS download center. An example of such a link would be https://aka.ms/vscode-dotnet-runtime/Runtime/8.0.0-rc.1.23419.4/dotnet-runtime-8.0.0-rc.1.23419.4-win-x64.zip - the link for the public RC1 stable version  of the 8.0.0-rc.1 .NET Runtime. We would only make links for the stable, released public versions of the .NET runtime, asp.net runtime, windows desktop runtime, and SDK.  The install scripts would be used with the `-AzureFeed https://aka.ms/vscode-dotnet-runtime` parameter passed, and this feed was successfully used as the base of the URI construction. However, we do not have a productversion.txt alias defined - and doing so would be a maintenance burden, as the productVersion.txt file contents would need to be updated for every release for every version. Currently, we _must_ provide this file, because not defining the alias results in a 301 redirect to bing.com instead of a 404 - which means the install scripts treat the entire contents of the bing search page as the product version.

Therefore, we'd like the add the following behavior:  

* A new flag, `-SkipProductVersionCheck` should be added (`--skip-product-version-check` in GNU syntax) to the scripts, and if this flag is provided then the scripts should skip checking the productVersion.txt file, relying instead only on the version requested by the user. If that version is incorrect in anyway, the operation is allowed to fail.

YuliiaKovalova:
@baronfel , do we still need to keep this ticket?

[...]

PR #403, opened 2023-10-10
require productVersion.txt file lookups to be application/octet-streams

baronfel:
This is an alternative to #402 - instead of skipping the productversion check, we instead enforce that the file is of the expected content-type. If not, the productVersion is not viable and we fallback to the other ways of determining it. This unblocks the proposed redirect scheme described in #402.

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -658,17 +658,19 @@ function Get-Product-Version([string]$AzureFeed, [string]$SpecificVersion, [stri
         try {
             $productVersionResponse = GetHTTPResponse($productVersionTxtUrl)
 
-            if ($productVersionResponse.StatusCode -eq 200) {
-                $productVersion = $productVersionResponse.Content.ReadAsStringAsync().Result.Trim()
-                if ($productVersion -ne $SpecificVersion)
-                {
-                    Say "Using alternate version $productVersion found in $ProductVersionTxtURL"
-                }
-                return $productVersion
+            if ($productVersionResponse.StatusCode -ne 200) {
+                Say-Verbose "Got unsuccessful StatusCode '$($productVersionResponse.StatusCode)' when trying to get productVersion.txt at $productVersionTxtUrl."
+                continue;
             }
-            else {
-                Say-Verbose "Got StatusCode $($productVersionResponse.StatusCode) when trying to get productVersion.txt at $productVersionTxtUrl."
+            if ($productVersionResponse.Content.Headers.ContentType.MediaType -ne "application/octet-stream") {
+                Say-Verbose "Got unknown Content-Type '$($productVersionResponse.Content.Headers.ContentType)' when trying to get productVersion.txt at $productVersionTxtUrl."
+                continue;
             }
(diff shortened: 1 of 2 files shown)


src/dotnet-install.sh
```diff
@@ -747,19 +747,39 @@ get_specific_product_version() {
...
         then
-            specific_product_version=$(wget -qO- "${download_link}${feed_credential}" 2>&1)
-            if [ $? = 0 ]; then
+            # wget is less friendly than curl to inspecting the response. We have to do a HEAD request, save the headers, and then search for the content-type header using grep.
+            # This gets us a value like ' Content-Type: text/html; charset=utf-8', so we have to
+            # 1. remove all leading whitespace
+            # 2. split on : and ; to create a space-delimited set of tokens
+            # 3. take the second token, which is the content type we want to compare against
+            local productVersionContentType=""
+            productVersionContentType=$(wget -qO- --method HEAD --save-headers --server-response --output-document - "${download_link}${feed_credential}" 2>&1 | grep 'Content-Type' | tr -d '[:space:]' | tr ':' ' ' | tr ';' ' ' | cut -d' ' -f 2)
```
ladipro, line 777: nit: I think some of these `tr` invocations can be eliminated. For example `tr ':;' ' '` instead of `tr ':' ' ' | tr ';' ' '`.

src/dotnet-install.ps1
```diff
@@ -658,17 +658,19 @@ function Get-Product-Version([string]$AzureFeed, [string]$SpecificVersion, [stri
...
-            if ($productVersionResponse.StatusCode -eq 200) {
-                $productVersion = $productVersionResponse.Content.ReadAsStringAsync().Result.Trim()
-                if ($productVersion -ne $SpecificVersion)
-                {
-                    Say "Using alternate version $productVersion found in $ProductVersionTxtURL"
-                }
-                return $productVersion
+            if ($productVersionResponse.StatusCode -ne 200) {
+                Say-Verbose "Got unsuccessful StatusCode '$($productVersionResponse.StatusCode)' when trying to get productVersion.txt at $productVersionTxtUrl."
+                continue;
```
ladipro, line 663: The semicolon should be removed, same for the other `continue` a few lines below.

ladipro (commented):
It's unusual for text files to be served as `application/octet-stream`. Should we get a sign-off from the folks running the server that this is unlikely to change?

YuliiaKovalova:
> It's unusual for text files to be served as `application/octet-stream`. Should we get a sign-off from the folks running the server that this is unlikely to change?

I think Ladi caught an issue here.
For failed tests productVersion.txt returns: _Content-Type text/plain; charset=utf-8._ Please see logs attached 
[ErrorLogs.txt](https://github.com/dotnet/install-scripts/files/12868412/ErrorLogs.txt)
[SuccessfulLogs.txt](https://github.com/dotnet/install-scripts/files/12868413/SuccessfulLogs.txt)

YuliiaKovalova:
Agreed to close this PR offline

Issue #406, opened 2023-10-24
--os should support macos


danmoseley:
OSX name changed to macOS in 2016. Even if existing scripts keep using osx, we should support macos as well. Otherwise we look rather out of the loop.

```
dan@danmose50:/mnt/c/WINDOWS/system32$ /mnt/c/git/install-scripts/src/dotnet-install.sh  -DryRun -Version 8.0.100-rtm.23523.2 --os mac
dotnet_install: Error: 'mac' is not a supported value for --os option, supported values are: osx, linux, linux-musl, freebsd, rhel.6. If you think this is a bug, report it at https://github.com/dotnet/install-scripts/issues.
dan@danmose50:/mnt/c/WINDOWS/system32$ /mnt/c/git/install-scripts/src/dotnet-install.sh  -DryRun -Version 8.0.100-rtm.23523.2 --os osx
```

baronfel:
Totally fair - we did this change in the RID graph a while back as well so should take it here to be in sync.

baronfel:
Closing this as we've merged this and deployed it to the .NET Website.

PR #408, opened 2023-10-27
Added macos as -os option

maridematte:
Issue https://github.com/dotnet/install-scripts/issues/406

Added the option for `macos` for `-os`. 

When `macos` is passed as an option, it is converted to `osx` so the download links still work.


baronfel (approved):
Thanks!

src/dotnet-install.sh
```diff
@@ -423,8 +423,13 @@ get_normalized_os() {
...
                 return 0
                 ;;
+            macos)
+                osname='osx'
+                echo "$osname"
+                return 0
+                ;;
             *)
-                say_err "'$user_defined_os' is not a supported value for --os option, supported values are: osx, linux, linux-musl, freebsd, rhel.6. If you think this is a bug, report it at https://github.com/dotnet/install-scripts/issues."
+                say_err "'$user_defined_os' is not a supported value for --os option, supported values are: osx, macos, linux, linux-musl, freebsd, rhel.6. If you think this is a bug, report it at https://github.com/dotnet/install-scripts/issues."
```
danmoseley, line 432:
```suggestion
                say_err "'$user_defined_os' is not a supported value for --os option, supported values are: osx, macos, linux, linux-musl, freebsd, rhel.6. If you think this is a bug, we would welcome a report in our repo at https://github.com/dotnet/install-scripts/issues."
```

danmoseley: Feel free to ignore! I am all about being friendly!

tests/Install-Scripts.Test/GivenThatIWantToGetTheSdkLinksFromAScript.cs
```diff
@@ -371,6 +371,29 @@ public async Task WhenAnExactVersionIsPassedToBash(string version, string runtim
...
+                    "--os", "macos",
+                    "-dryrun" };
+
+            var commandResult = CreateInstallCommand(args)
+                            .CaptureStdOut()
+                            .CaptureStdErr()
+                            .Execute();
+
+            commandResult.Should().Pass();
+            commandResult.Should().NotHaveStdErr();
```
danmoseley, line 393: Verify it used osx?

danmoseley (approved):

tests/Install-Scripts.Test/GivenThatIWantToGetTheSdkLinksFromAScript.cs
danmoseley: ah I see you do

maridematte: Yes, I added on a recent commit

YuliiaKovalova:
Looks good.
Please extend the documentation too: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script#:~:text=%2D%2D-,os,-%3COPERATING_SYSTEM%3E

YuliiaKovalova (approved):

PR #409, opened 2023-11-07
Add usage examples of the two primary modes of invoking this script

baronfel:
Also provide a user shorthand that maps `-help` to `Get-Help -Examples` based on recurring user feedback.

We get a lot of internal feedback about the usability of the scripts - this is a small change to provide some usage-based examples of how to use the script, as well as provide a shortcut for a common user gesture for `-help`. Many folks that work on multiple platforms aren't used to reaching for `Get-Help` in Powershell, so we can do that for them.

I also tweaked the bash usage examples to show the bi-modal nature of the script.


src/dotnet-install.ps1
```diff
@@ -1170,6 +1177,12 @@ function Prepare-Install-Directory {
     }
 }
 
+if ($Help)
+{
+    Get-Help $PSCommandPath -Examples
+    exit
+}
```
f-alizada, line 1184: Thank you @baronfel for the PR! Will it make sense to cover those cases by tests to verify that the output is the one that we expect? Also test coverage will allow us to catch changes of help method when they are not expected :) 
What do you think?

baronfel: I added some basic tests - we don't have any coverage of this currently so this will at least pin the current state. I don't know why the jobs got stuck, though.

f-alizada: Thank you! Currently there is an ongoing PR: https://github.com/dotnet/install-scripts/pull/411 with the updated tests which should solve this.

f-alizada: I have updated the tests. 
the problem mainly was in the verify command, since it requires to have something to compare with which was not added at that point. However I changed the approach to use explicit comparison of some rows of the ouput because the all help output could differ from machine to machine depending on the configuration.

baronfel: Thank you @f-alizada!

baronfel:
I'm happy to merge this as soon as someone approves it.

YuliiaKovalova (approved):

PR #413, opened 2023-11-16
Tests updates after release dotnet 8

f-alizada:
## Problem
After recent LTS version upgrade current tests are failing. 

## Solution
The PR is updating current state of some tests which involves downloading LTS version of sdk or specific runtimes. 
It is also a prerequisite for the https://github.com/dotnet/install-scripts/pull/411 to complete.

### DIFF
--- a/tests/Install-Scripts.Test/GivenThatIWantToInstallDotnetFromAScript.cs
+++ b/tests/Install-Scripts.Test/GivenThatIWantToInstallDotnetFromAScript.cs
@@ -33,7 +33,7 @@ namespace Microsoft.DotNet.InstallationScript.Tests
                 ("6.0", "6\\.0\\..*", Quality.Daily),
                 ("6.0", "6\\.0\\..*", Quality.None),
                 ("STS", "7\\.0\\..*", Quality.None),
-                ("LTS", "6\\.0\\..*", Quality.None),
+                ("LTS", "8\\.0\\..*", Quality.None),
                 ("7.0", "7\\.0\\..*", Quality.None),
                 ("7.0", "7\\.0\\..*", Quality.Ga),
                 ("8.0", "8\\.0\\..*", Quality.None),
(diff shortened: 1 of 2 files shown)


YuliiaKovalova (approved):

tests/Install-Scripts.Test/GivenThatIWantToInstallDotnetFromAScript.cs
```diff
@@ -33,9 +33,11 @@ public class GivenThatIWantToInstallDotnetFromAScript : IDisposable
                 ("6.0", "6\\.0\\..*", Quality.Daily),
                 ("6.0", "6\\.0\\..*", Quality.None),
                 ("STS", "7\\.0\\..*", Quality.None),
-                ("LTS", "6\\.0\\..*", Quality.None),
```
YuliiaKovalova, line 36: This test is relevant,  please leave it as it is and add a separate for 8.0.xx

f-alizada: Thank you for the review! 
May I clarify why the test is relevant since the latest Long Term Support version is 8? 
Channel used for the test is LTS, according to the docs: https://learn.microsoft.com/en-gb/dotnet/core/tools/dotnet-install-script#options we are downloading the most recent version of long term support. 
Am I missing something?

YuliiaKovalova: You can check the support of the versions on this page https://dotnet.microsoft.com/en-us/download/dotnet
As you can see, net 6.0 saves it's status "LTS". It happens in parallel to net 8.0 "LTS" marker :)

f-alizada: Thank you for checking it! 
The status of net 6.0 is LTS and will be supported however it is not recent one, aka.ms links are updated to download the dotnet 8 version, since it is the recent one, sorry for the confusions if I'm missing something. 
Also without updating the LTS regular expression to 8 the tests autoamtically fail since it downloads the dotnet 8.

YuliiaKovalova: Oh, I see, you are right, sorry for misleading you.

f-alizada: Thank you for raising the question! resolving the conversation

PR #435, opened 2024-02-03
Detect ARMhf userland on top of AARCH64 kernel

filipnavara:
Raspberry Pi 4/5 supports running 32-bit user land with a 64-bit Linux kernel. Detect that and install linux-arm version of .NET instead of linux-arm64. The arm64 version doesn't work in that scenario.


YuliiaKovalova:
Looks good to me!
@baronfel , please review too

PR #441, opened 2024-02-19
Update install-dotnet-preview.sh to support install of 9.0.0-preview

rodgfr:
The current script is linked in https://github.com/dotnet/core/blob/main/release-notes/9.0/install-linux.md but does not support the installation of 9.0.0-preview. This update the links to install the current .net 9 preview.


leecow (approved):
LGTM

PR #451, opened 2024-03-14
1ES Templates onboarding

YuliiaKovalova:
Split the pipeline on production and PR one.
Onboard 1es official templates for prod yml, validated internally https://dev.azure.com/dnceng/internal/_build/results?buildId=2404188&view=results

On azure devops side everything is ready for new naming.


azure-pipelines-PR.yml
```diff
@@ -0,0 +1,87 @@
...
+          displayName: Run Tests
+          inputs: 
+            command: test
+
+  - job: LinuxTests
+    timeoutInMinutes: 30
+    container: UbuntuContainer
+    pool:
+      name: NetCore-Public
+      demands: ImageOverride -equals Build.Ubuntu.1804.Amd64.Open
```
baronfel, line 52: this is a pretty old version of Ubuntu - at what point should we update?

rainersigwald: Follow-up question: _should_ this be running tests on like a min-spec OS?

azure-pipelines-official.yml
```diff
@@ -0,0 +1,162 @@
...
+        - task: DotNetCoreCLI@2
+          displayName: Run Tests
+          inputs: 
+            command: test
+
+      - job: MacOSTests
+        timeoutInMinutes: 30
+        pool: 
+          name: Azure Pipelines
+          image: macOS-11
```
rainersigwald, line 120: Similarly this is pretty old, _should_ it be?

rainersigwald (approved):

azure-pipelines-PR.yml
YuliiaKovalova: looking at vstest pipelines state, the image selection seems to be reasonable: 
https://github.com/microsoft/vstest/blob/main/azure-pipelines.yml

azure-pipelines-official.yml
YuliiaKovalova: I have double checked with the 1es team - it's ok to bump it.

baronfel (approved):

PR #464, opened 2024-04-25
Correct broken redirects in the case that the response ends without final code 2xx/4xx/5xx

### LINKED ISSUE #461: dotnet 9 install fails due to aka.ms resolving to bing
Don't think this issue is directly the scripts fault, but seemed like the best place to report it.

When using dotnet-install.sh to install `--channel 8.0` and prior, the install link correctly resolves to azureedge.net and the install completes fine:

```bash
dotnet-install: Attempting to download using primary link https://dotnetcli.azureedge.net/dotnet/Sdk/8.0.202/dotnet-sdk-8.0.202-linux-x64.tar.gz
```
but when attempting to install `--channel 9.0` it instead resolves to a bing url which fails.

```bash
dotnet-install: Attempting to download using aka.ms link https://www.bing.com?ref=aka&shorturl=dotnet/9.0/dotnet-sdk-linux-x64.tar.gz
curl: (7) Failed to connect to www.bing.com port 443 after 0 ms: Connection refused
```
Now the failure is likely due to our network having restrictions on bing.com, but it seemed a little odd that the aka.ms resolves to bing rather than directly to azureedge.net like all the other download pages do. Any chance this could be fixed by whomever is responsible?

GangWang01:
### Fix
#461

### Solution
In the case that the response ends without final code 2xx/4xx/5xx, getting broken redirects includes last http code.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -1271,6 +1271,12 @@ get_download_link_from_aka_ms() {
     http_codes=$( echo "$response" | awk '$1 ~ /^HTTP/ {print $2}' )
     # They all need to be 301, otherwise some links are broken (except for the last, which is not a redirect but 200 or 404).
     broken_redirects=$( echo "$http_codes" | sed '$d' | grep -v '301' )
+    # The response may end without final code 2xx/4xx/5xx somehow, e.g. network restrictions on www.bing.com causes redirecting to bing.com fails with connection refused.
+    # In this case it should not exclude the last.
+    last_http_code=$(  echo "$http_codes" | tail -n 1 )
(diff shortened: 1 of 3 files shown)


YuliiaKovalova:
Could you please cover the case with .net 9 here and check other places?
https://github.com/GangWang01/install-scripts/blob/cc56933f989275e5dabe35441e224845ef9e7a44/tests/Install-Scripts.Test/GivenThatIWantToInstallDotnetFromAScript.cs#L40

YuliiaKovalova (approved):

YuliiaKovalova:
thank you!

PR #497, opened 2024-07-29
detect IBM Power architecture

barf:
IBM Power architecture is not correctly detected on FreeBSD 15.0-CURRENT

this patch is one step towards the bringup on this arch

but the script fails because Microsoft don't provide any pre-built PPC dotnet sdk (another issue)

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -331,6 +331,10 @@ get_machine_architecture() {
             echo "riscv64"
             return 0
             ;;
+        powerpc|ppc)
+            echo "ppc"
+            return 0
+            ;;
         esac
     fi
 


f-alizada:
Hello @barf thank you for the PR! 
Having that the script will still be failing, I would suggest to wait  untill dependent issue is solved/closed.
> but the script fails because Microsoft don't provide any pre-built PPC dotnet sdk (another issue)

cc: @baronfel

baronfel:
Thanks for this - it seems perfectly reasonable to extend the current arch matching to accept this. The existing portable powerpc arch is `ppc64le` - does that work for your case?

MS likely isn't going to provide any pre-built PowerPC binaries - one of the roles of this script is to allow third parties that _do_ want to make these binaries to download and manage them in the same way as the MS-authored binaries, via use of the custom feed url parameters.

Just to check myself, @leecow are there any plans for PowerPC support for Microsoft-build SDKs and Runtimes?

barf:
> Thanks for this - it seems perfectly reasonable to extend the current arch matching to accept this. The existing portable powerpc arch is `ppc64le` - does that work for your case?

the little endian power machines are newer than what I have (Apple Xserve G5 and IBM p5), so no ppc64le is not the correct architecture but AFAIK should be detected OK.

> MS likely isn't going to provide any pre-built PowerPC binaries - one of the roles of this script is to allow third parties that _do_ want to make these binaries to download and manage them in the same way as the MS-authored binaries, via use of the custom feed url parameters.

appreciate the situation and expect to patch/build the runtime myself if I can, i do not expect MS support for this, the situation for users of C# is just going backwards because PPC and PPC64 support exists in Mono, but not here in the actual dot net. i just need to run an application made with framework 8 but only have PPC hardware, if it was made in 4.7 its no problem, if that helps explain?

C# and a good chunk of the framework (via Mono) was already running on old Apple PPC machines, unfortunately the whole MS sponsorship of Mono project has not been conducive to extending this support which brings me here :-)

baronfel:
You might have the wrong idea about this repo - it's explicitly not for downloading NET Framework or legacy Mono bits at all.

barf:
> You might have the wrong idea about this repo - it's explicitly not for downloading NET Framework or legacy Mono bits at all.

thanks but i think i get that. the context here is (like ARM until recently) Mono had better C# support for PPC users and dotnet needs to improve platform support to get to that point Mono was at. Mono has done the PPC JIT work, but the new dotnet runtime is only targeting ppc64le, it seems achievable to merge or bring forward some of that work.

but there is obviously more to patching some download scripts to bring up on this arch, im working on the cmake scripts on the runtime atm, will do a PR there when I have something more

baronfel:
Ok, gotcha - this is part of new platform bring-up. Sorry, the mentions of mono really threw me off. All clear now!

baronfel (approved):

barf:
> Ok, gotcha - this is part of new platform bring-up. Sorry, the mentions of mono really threw me off. All clear now!

sorry if I wasn't clear! yes one small step towards bring up on FreeBSD PPC & PPC64, also I have a PPC64 Debian lab machine that could be running dotnet!

leecow:
> Just to check myself, @leecow are there any plans for PowerPC support for Microsoft-build SDKs and Runtimes?

No plans for MS-built native support.

barf:
> > Just to check myself, @leecow are there any plans for PowerPC support for Microsoft-build SDKs and Runtimes?
> 
> No plans for MS-built native support.

thank you for clarifying, FYI FreeBSD distribute mono for PPC/PPC64 so we can run Roslyn on old Apple PPC machines already, i'm hoping MS don't ignore the interest in dotnet on PPC/PPC64 there are business applications for this!

PR #513, opened 2024-08-30
Check if user has write access to the install root

### LINKED ISSUE #496: Path assembly bug in dotnet-install.ps1
On line 770, `$InstallRoot` is set to the value of the `DOTNET_INSTALL_DIR` environment variable (`C:\Program Files\dotnet\sdk` in my case). This results in the creation of a duplicate `sdk` folder name in the `Is-Dotnet-Package-Installed()` function at line 789.

`$DotnetPackagePath` is incorrectly set to `C:\Program Files\dotnet\sdk\sdk\8.0.303`.

This results in the script attempting to proceed with the installation instead of failing gracefully (at line 1234), which in turn results in an inappropriate error:
```
dotnet-install: Failed to extract package. Exception: Exception calling "ExtractToFile" with "3" argument(s): "Access to the path 'C:\Program Files\dotnet\sdk\dotnet.exe' is denied."
Exception calling "ExtractToFile" with "3" argument(s): "Access to the path 'C:\Program Files\dotnet\sdk\dotnet.exe' is denied."
At E:\Users\Work\Downloads\dotnet-install.ps1:864 char:21
+ ...             [System.IO.Compression.ZipFileExtensions]::ExtractToFile( ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : UnauthorizedAccessException
```
I'm running the script non-elevated and with no parameters.

GangWang01:
### Fix 
#496

### Solution
Add the check that tests if user has write access to install root. If user doesn't have, fail the installation with an error telling running as administrator.
Additionally, if user specifies the install root via environment variable `DOTNET_INSTALL_DIR` to the sub folder of `$env:ProgramFiles\dotnet`, warn that it's better to keep aligned with .NET SDK installer.

### DIFF
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -1222,7 +1222,7 @@ Measure-Action "Product discovery" {
 
 $InstallRoot = Resolve-Installation-Path $InstallDir
 if (-not (Test-User-Write-Access $InstallRoot)) {
-    Say-Error "Current user doesn't have write access to the installation root $InstallRoot to install dotnet. Please try running as administrator to resolve."
+    Say-Error "The current user doesn't have write access to the installation root '$InstallRoot' to install .NET. Please try specifying a different installation directory using the -InstallDir parameter, or ensure the selected directory has the appropriate permissions."
     throw
 }
 Say-Verbose "InstallRoot: $InstallRoot"
--- a/src/dotnet-install.ps1
+++ b/src/dotnet-install.ps1
@@ -771,6 +771,9 @@ function Get-User-Share-Path() {
     if (!$InstallRoot) {
         $InstallRoot = "$env:LocalAppData\Microsoft\dotnet"
     }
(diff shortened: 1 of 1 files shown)


YuliiaKovalova:
@GangWang01 , thank you for handling this!
It is possible to cover this case with UTs?

nlsdg:
As the stated purpose of these scripts is "perform a non-admin installation", I think the message to the user shouldn't just be to try running as administrator. (Also: if they had admin access they could also just run a normal install.)
The possibility to use/change the -InstallDir parameter could be added to the message to make theuser aware it exists or might be wrong.

GangWang01:
> It is possible to cover this case with UTs?

Running in CI it's a little difficult to get a normal user with general permission for the test. I just did testing locally.

YuliiaKovalova (approved):

PR #526, opened 2024-10-08
Fix net10.0 detection for osx-arm64

am11:
It was downloading osx-x64 on osx-arm64 machine due to relaxed pattern matching:

```sh
$ ./dotnet-install.sh --quality daily --channel "10.0" --install-dir ~/.dotnet10 -verbose

dotnet-install: Calling: machine_has curl
dotnet-install: Calling: calculate_vars 
dotnet-install: Calling: get_normalized_architecture_from_architecture <auto>
dotnet-install: Calling: get_machine_architecture 
dotnet-install: Normalized architecture: 'arm64'.
dotnet-install: Calling: get_normalized_os 
dotnet-install: Calling: get_current_os_name 
dotnet-install: Normalized OS: 'osx'.
dotnet-install: Calling: get_normalized_quality daily
dotnet-install: Normalized quality: 'daily'.
dotnet-install: Calling: get_normalized_channel 10.0
dotnet-install: Normalized channel: '10.0'.
dotnet-install: Calling: get_normalized_product 
dotnet-install: Normalized product: 'dotnet-sdk'.
dotnet-install: Calling: resolve_installation_path /Users/adeel/.dotnet10
dotnet-install: InstallRoot: '/Users/adeel/.dotnet10'.
dotnet-install: Calling: get_normalized_architecture_for_specific_sdk_version Latest 10.0 arm64
dotnet-install: Calling: get_current_os_name 
dotnet-install: Changing user architecture from 'arm64' to 'x64' because .NET SDKs prior to version 6.0 do not support arm64.
...
```

This delta fixes the issue by parsing major version from the input and comparing if it is less than 6.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -423,14 +423,16 @@ get_normalized_architecture_for_specific_sdk_version() {
 # args:
 # version or channel - $1
 is_arm64_supported() {
-    #any channel or version that starts with the specified versions
-    case "$1" in
-        ( "1"* | "2"* | "3"*  | "4"* | "5"*) 
-            echo false
-            return 0
-    esac
+    # Extract the major version by splitting on the dot
+    major_version="${1%%.*}"
 
-    echo true
+    # Check if the major version is a valid number and less than 6
+    if [[ "$major_version" -lt 6 ]]; then
+        echo false
+    else
+        echo true
+    fi
+
     return 0
 }
 
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -427,12 +427,16 @@ is_arm64_supported() {
     major_version="${1%%.*}"
 
     # Check if the major version is a valid number and less than 6
-    if [[ "$major_version" -lt 6 ]]; then
-        echo false
-    else
-        echo true
-    fi
+    case "$major_version" in
+        [0-9]*)  
+            if [ "$major_version" -lt 6 ]; then
+                echo false
+                return 0
+            fi
+            ;;
+    esac
 
+    echo true
     return 0
 }
 


am11:
@baronfel, could we please deploy this to dot.net after the merge? I’m concerned that as people start updating to the 10.0 previews, they may run into some unexpected issues on osx-arm64. I discovered this myself (after a cup of ☕) because the use of Rosetta emulation isn't clearly highlighted, which might lead to confusion about the fact that .NET is actually emulating. 😅 Thanks!

baronfel:
I've asked the team to review, and once we merge we'll definitely work with @mairaw's team to get it updated 👍

MichalPavlik (approved):

MichalPavlik:
I'm not a Bash person, but it looks good. I would like to add one small change to cover another issue and then I can start the deployment process. It takes some time as we need to test the new version in 4 major .NET repos before publishing to dot.net. But I think it could be done this week.

am11:
Thank you! 🎉

MichalPavlik:
Thanks for the PR @am11 :)

src/dotnet-install.sh
```diff
@@ -423,11 +423,17 @@ get_normalized_architecture_for_specific_sdk_version() {
 # args:
 # version or channel - $1
 is_arm64_supported() {
-    #any channel or version that starts with the specified versions
-    case "$1" in
-        ( "1"* | "2"* | "3"*  | "4"* | "5"*) 
-            echo false
-            return 0
+    # Extract the major version by splitting on the dot
+    major_version="${1%%.*}"
```
rolfbjarne, line 427: I guess this will have to be updated again in 10 years, for .NET 20 :smile:

am11: Haha, I see what you're saying! But no worries, since we're already checking for versions less than six, the script will handle .NET 20 and beyond just fine.

I think there might be a bit of confusion about `${1}`, it’s actually referring to the first positional argument passed into the function, not the number `1` as the major version. To make things clearer, maybe we could assign `$1` to a more descriptive variable like `version_string` and then extract the major version from there. Something like:

```bash
version_string="$1"
major_version="${version_string%%.*}"
```

Just a thought for better readability! 😊

rolfbjarne: > > I guess this will have to be updated again in 10 years, for .NET 20 😄
> 
> the 1 in ${1%%.*} is arg number not the version standard shell syntax
> 

definitely not my sharpest moment... 🤦‍♂️

PR #534, opened 2024-10-19
Avoid duplicate errors

kasperk81:
when an "attempt" fails, it shows:

``` 
dotnet-install: Attempting to download using primary link ...
curl: (56) The requested URL returned error: 404
dotnet-install: The resource at primary link '...' is not available.
```

the middle line is redundant with `--verbose` and isn't part of the expected output in regular usage. removing it to avoid being flagged as an error by systems like [CanonicalError](https://github.com/dotnet/msbuild/blob/main/src/Shared/CanonicalError.cs#L243).


baronfel:
I'm a little concerned about not seeing the curl errors at all - there are some error-handling routines in the curl/wget functions that return specific error messages, but sometimes a status code isn't enough to provide context for investigation. If we swallow all of these errors, how would users get detailed diagnostics? Is there maybe a way to preserve these error logs for later investigation?

Issue #535, opened 2024-10-22
cp: unrecognized option: n


jurschel:
This morning all of my pipelines stopped working and are failing with this error.
cp: unrecognized option: n
I believe this is coming from the dotnet-install.sh script that is pulled down at runtime when my pipelines run.  At the moment all of the pipelines are broken due to this.

```$ ./build.sh Version
GNU bash, version 5.1.16(1)-release (x86_64-alpine-linux-musl)
dotnet-install: Attempting to download using primary link https://dotnetcli.azureedge.net/dotnet/Sdk/3.1.426/dotnet-sdk-3.1.426-linux-musl-x64.tar.gz
dotnet-install: Remote file https://dotnetcli.azureedge.net/dotnet/Sdk/3.1.426/dotnet-sdk-3.1.426-linux-musl-x64.tar.gz size is 123630520 bytes.
dotnet-install: Extracting archive from https://dotnetcli.azureedge.net/dotnet/Sdk/3.1.426/dotnet-sdk-3.1.426-linux-musl-x64.tar.gz
cp: unrecognized option: n
BusyBox v1.[33](https://xxx.com/xxx/-/jobs/8153870880#L33).1 () multi-call binary.
Usage: cp [-arPLHpfilsTu] SOURCE... DEST
Copy SOURCE(s) to DEST
	-a	Same as -dpR
	-R,-r	Recurse
	-d,-P	Preserve symlinks (default if -R)
	-L	Follow all symlinks
	-H	Follow symlinks on command line
	-p	Preserve file attributes if possible
	-f	Overwrite
	-i	Prompt before overwrite
	-l,-s	Create (sym)links
	-T	Treat DEST as a normal file
	-u	Copy only newer files

**Comment by baronfel** [comment]
Can you provide more information about 
* where you are getting the scripts from precisely (the URL?)
* your OS/runtime environment?
* the version of the `cp` command that you have available on your system?

**Comment by rootwyrm** [comment]
@baronfel  let's go ahead and mark this as a confirmed bug. I discussed with @jurschel and the precise reproduction is on `docker:20.10.11` which is alpine 3.14.3, which previously worked.

A super quick reproduction would be:
```
docker run --rm -it docker:20.10.11 /bin/sh
apk add --no-cache -q curl bash
curl -o dotnet-install.sh https://dotnet-install.sh ; chmod +x ./dotnet-install.sh
./install.sh --channel 3.1 --no-path --verbose
```

[...]

PR #536, opened 2024-10-22
Make cp -u check robust

am11:
Fix https://github.com/dotnet/install-scripts/issues/535.

@rootwyrm thanks for the additional info. Could you please give it a try? I've tested it on Alpine with BusyBox v1.36.1.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -957,6 +957,37 @@ get_absolute_path() {
 }
 
 # args:
+# override - $1 (boolean, true or false)
+get_cp_options() {
+    eval $invocation
+
+    local override="$1"
+    local override_switch=""
+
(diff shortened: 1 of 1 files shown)


JanProvaznik (changes requested):
Thanks for the fix, could you please extract the -u / -n availability checking to a function?

JanProvaznik:
I tested it on BusyBox v1.33.1, it works

rootwyrm:
+1 to @JanProvaznik on both items. Succeeding for all previously failing cases here too. :)

am11:
Thanks for the confirmation guys. Super glad that there was a way forward (was afraid of rolling it back now that Ubuntu builds are getting clean logs). 😅

JanProvaznik (approved):

PR #562, opened 2025-01-02
Fix spelling in error string

cinderblocks:
Fix spelling in string "Cound not construct a legacy_download_link; omitting..."


JanProvaznik (approved):
thanks 👍

PR #585, opened 2025-03-28
Update tests and signing to target net9.0

JanProvaznik:
note: needs arcade 9


JanProvaznik:
this upgrade is not really helpful, it can be done with .net 10 lts

Issue #610, opened 2025-06-02
dotnet-install.sh fails with curl 8.14.0


mchaves27:
Issue Summary: When installing .NET using the dotnet-install.sh script, the installation fails with the error:
`/usr/local/bin/dotnet-install.sh: line 1380: link_types[$link_index]: unbound variable`
This issue started occurring after upgrading `curl` from version `8.13.0` to `8.14.0`.

Environment:
Base Image: Red Hat Enterprise Linux (RHEL) 8
Shell: /bin/sh (within Docker build)
dotnet-install.sh  Source: Private GitHub Proxy Repository
Previous curl Version (Working): 8.13.0
Current curl Version (Failing): 8.14.0

Steps to reproduce: 
Build a Docker image using RHEL 8 as the base.
Add custom `curl`  and `dotnet` installation:

```docker
RUN CURL_SOURCE_URL=$(curl -sSL https://api.github.com/repos/curl/curl/releases/latest | jq -r '.assets[] | select(.name | endswith(".tar.gz")) | .browser_download_url' | sed 's/github\.com/some-private-repository\/repository\/github-proxy/') \
  && curl -sSL $CURL_SOURCE_URL -o curl.tar.gz \
  && tar -xzf curl.tar.gz \
  && rm -f curl.tar.gz \
  && CURL_LOCATION=$(ls -d curl-*/) \
  && pushd ${CURL_LOCATION} \
  && ./configure \
  --enable-static --disable-shared \
  --with-openssl --with-brotli --with-zstd \
  --with-nghttp2 \
  --with-libidn2 --with-libssh2 \
  --enable-hsts --enable-mime --enable-cookies \
  --enable-http-auth --enable-manual \
  --enable-proxy --enable-file --enable-http \
  --enable-ftp --enable-telnet --enable-tftp \
  --enable-pop3 --enable-imap --enable-smtp \
  --enable-gopher --enable-mqtt \
  --enable-doh --enable-dateparse --enable-verbose \
  --enable-alt-svc --enable-websockets \
  --enable-ipv6 --enable-unix-sockets --enable-socketpair \
  --enable-headers-api --enable-versioned-symbols \
  --enable-optimize --enable-pthreads \
  --enable-warnings --enable-werror \
  --enable-curldebug --enable-dict --enable-netrc \
  --enable-bearer-auth --enable-tls-srp --enable-dnsshuffle \
  --enable-get-easy-options --enable-progress-meter \
[...]

PR #617, opened 2025-06-12
Handle curl 8.14.0 zero-exit bug

ellahathaway:
Fixes https://github.com/dotnet/install-scripts/issues/610
Related to https://github.com/curl/curl/issues/17554.

Curl 8.14.0 was returning a zero-exit code even when the output indicated a failure, such as curl: (22) The requested URL returned error: 404. This behavior resulted in failures as described in https://github.com/dotnet/install-scripts/issues/610.

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -1208,7 +1208,7 @@ downloadcurl() {
 
     # Regression in curl causes curl with --retry to return a 0 exit code even when it fails to download a file - https://github.com/curl/curl/issues/17554
     if [ $curl_exit_code -eq 0 ] && echo "$curl_output" | grep -q "^curl: ([0-9]*) "; then
-        curl_exit_code=$(echo "$curl_output" | grep "^curl: ([0-9]*) " | sed 's/curl: (\([0-9]*\)).*/\1/')
+        curl_exit_code=$(echo "$curl_output" | sed 's/curl: (\([0-9]*\)).*/\1/')
     fi
 
     if [ $curl_exit_code -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


src/dotnet-install.sh
```diff
@@ -1198,13 +1198,19 @@ downloadcurl() {
...
     else
-        curl $curl_options -o "$out_path" "$remote_path_with_credential" 2>&1
+        curl_output=$(curl $curl_options -o "$out_path" "$remote_path_with_credential" 2>&1)
         curl_exit_code=$?
     fi
-    
+
+    # Regression in curl causes curl with --retry to return a 0 exit code even when it fails to download a file - https://github.com/curl/curl/issues/17554
+    if [ $curl_exit_code -eq 0 ] && echo "$curl_output" | grep -q "^curl: ([0-9]*) "; then
+        curl_exit_code=$(echo "$curl_output" | grep "^curl: ([0-9]*) " | sed 's/curl: (\([0-9]*\)).*/\1/')
```
NikolaMilosavljevic, line 1211: Could this be simplified to?
```
curl_exit_code=$(echo "$curl_output" | sed 's/curl: (\([0-9]*\)).*/\1/')
```

src/dotnet-install.sh
```diff
@@ -1198,13 +1198,19 @@ downloadcurl() {
...
     local curl_exit_code=0;
     if [ -z "$out_path" ]; then
-        curl $curl_options "$remote_path_with_credential" 2>&1
+        curl_output=$(curl $curl_options "$remote_path_with_credential" 2>&1)
         curl_exit_code=$?
+        echo "$curl_output"
     else
-        curl $curl_options -o "$out_path" "$remote_path_with_credential" 2>&1
+        curl_output=$(curl $curl_options -o "$out_path" "$remote_path_with_credential" 2>&1)
         curl_exit_code=$?
```
NikolaMilosavljevic, line 1206: Presumably we do not care for losing the progress status output?

src/dotnet-install.sh
ellahathaway: Addressed in https://github.com/dotnet/install-scripts/pull/617/commits/9b0e0600f7faeba21a74db04a1bffec587c77107

PR #693, opened 2026-03-20
Validate channel format in dotnet-install.sh

### LINKED ISSUE #686: Improve output for unsupported one-part channel in dotnet-install.sh
When a one-part channel (e.g. `-c 10`) is specified for dotnet-install.sh, the script attempts to resolve it and fails with unclear messaging. The output currently attempts to download files for the one-part channel, leading to a confusing 404 error. The script help text specifies that channels must be 'STS', 'LTS', a two-part version (A.B), or a three-part SDK band (A.B.Cxx), but this expectation isn't reflected in the error message.

Request:
- Detect if a one-part channel is given and update the output to indicate it is not valid.
- Make the error message clear about supported channel formats and provide a suggested workaround (e.g. use `-c 10.0`).

This will help users avoid confusion and parallels a similar situation described in issue #685 (and previously #461).

Repro:
```bash
bash dotnet-install.sh -c 10 --verbose
# output is unclear, attempted downloads fail with 404
```

Expected:
- Clear error about channel format and workaround.

luissena:
## Summary

Fixes #686

When an unsupported channel format is provided (e.g., `-c 10`), the script now exits early with a clear error message instead of proceeding to construct invalid URLs that fail with confusing 404 errors.

### Changes

- **`src/dotnet-install.sh`**: Added channel format validation in `get_normalized_channel()`. The wildcard case now validates that the channel matches either a 2-part version (`A.B`, e.g., `10.0`) or a 3-part SDK band version (`A.B.Cxx`, e.g., `5.0.1xx`). Invalid formats produce a clear error message listing supported formats, following the same pattern used by `get_normalized_quality()`.

- **`tests/.../GivenThatIWantToGetTheSdkLinksFromAScript.cs`**: Added `[InlineData("10")]` test case and updated the expected error message assertion to match the new validation message.

### Before

```
$ bash dotnet-install.sh -c 10 --verbose
# ... attempts to download from invalid URLs, gets 404
# "Failed to resolve the exact version number."
```

### After

```
$ bash dotnet-install.sh -c 10
# "'10' is not a supported value for --channel option. Supported values are:
#  STS, LTS, a 2-part version in A.B format (e.g., 10.0), or a 3-part SDK
#  band version in A.B.Cxx format (e.g., 5.0.1xx)."
```

### DIFF
--- a/src/dotnet-install.sh
+++ b/src/dotnet-install.sh
@@ -524,8 +524,16 @@ get_normalized_channel() {
                 return 0
                 ;;
             *)
-                echo "$channel"
-                return 0
+                if [[ "$channel" =~ ^[0-9]+\.[0-9]+$ ]]; then
+                    echo "$channel"
+                    return 0
+                elif [[ "$channel" =~ ^[0-9]+\.[0-9]+\.[0-9]+[a-z]{2}$ ]]; then
+                    echo "$channel"
+                    return 0
+                else
+                    say_err "'$1' is not a supported value for --channel option. Supported values are: STS, LTS, a 2-part version in A.B format (e.g., 10.0), or a 3-part SDK band version in A.B.Cxx format (e.g., 5.0.1xx). If you think this is a bug, report it at https://github.com/dotnet/install-scripts/issues."
+                    return 1
+                fi
                 ;;
         esac
     fi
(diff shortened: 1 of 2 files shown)


src/dotnet-install.sh
```diff
@@ -524,8 +524,16 @@ get_normalized_channel() {
                 return 0
                 ;;
             *)
-                echo "$channel"
-                return 0
+                if [[ "$channel" =~ ^[0-9]+\.[0-9]+$ ]]; then
+                    echo "$channel"
+                    return 0
+                elif [[ "$channel" =~ ^[0-9]+\.[0-9]+\.[0-9]+[a-z]{2}$ ]]; then
```
mmitche, line 530: I think this needs to be a bit more flexible and consider daily and preview packages. They'll have channel names like:

11.0.1xx-preview2:

`.\.dotnet\dotnet-install.ps1 -channel 11.0.1xx-preview2 -verbose -Quality daily`

PR #717, opened 2026-05-11
Use arcade signing

### LINKED ISSUE #716: Signing failing after upgrade to .NET 10
After upgrading to .NET 10 in https://github.com/dotnet/install-scripts/pull/708, signing the scripts on windows is failing with the following error:

```
##[error]C:\Users\cloudtest\.nuget\packages\microsoft.dotnet.arcade.sdk\10.0.0-beta.26222.2\tools\Publish.proj(428,5): error : (NETCORE_ENGINEERING_TELEMETRY=Publish) ItemsToPush is not specified.
```

[Pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2971244&view=logs&j=0bc77094-9fcd-5c38-f6e4-27d2ae131589&t=a78ba58e-354d-5443-f28e-fcee6d8f5be6) (internal Microsoft link)

ellahathaway:
- Fixes https://github.com/dotnet/install-scripts/issues/716
- Adds test signing to PR validation
- Moves to arcade's SignTool for signing

[Internal pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2972389&view=results)


eng/common-variables.yml
```diff
@@ -8,18 +8,24 @@ variables:
     value: True
   - name: _RunAsInternal
     value: False
-  - name: _InternalBuildArgs
-    value: ''
   - name: _UseBuildManifest
     value: False
   - name: _SignType
     value: test
+  - name: _BuildConfig
```
MichaelSimons, line 15: Curious, is there a particular reason to build w/debug in public/prs?  Does this provide value?

azure-pipelines-official.yml
```diff
@@ -148,7 +137,7 @@ extends:
         - checkout: self
           displayName: Checkout Self
           clean: true
-        - script: ./build.sh --test
+        - script: ./build.sh --test -configuration $(_BuildConfig)
```
MichaelSimons: Should we collect binlogs here are well?

azure-pipelines-PR.yml
```diff
@@ -13,6 +13,12 @@ stages:
   jobs:
   - template: /eng/common/templates/jobs/jobs.yml
     parameters:
+      artifacts:
+        publish:
+          artifacts: true
```
MichaelSimons, line 18: IIRC setting both artifacts and logs to true cause the logs to be published twice in separate artifact buckets.  See Artifacts_Windots_NT_Debug_Attempt1 and Logs_Build_Windows_NT_Debug in the PR validation.  One or the other should be used at a time.  It is definitely a poor UX.

MichaelSimons (approved):

azure-pipelines-official.yml
ellahathaway: In my opinion, I only see them being useful for investigations into the signing process.

eng/common-variables.yml
ellahathaway: Not that I know of. But I agree that building in with release is more reflective of the internal CI.

azure-pipelines-PR.yml
ellahathaway: You're right. `Artifacts_Windots_NT_Debug_Attempt1` is the correct one in this case. I'll fix.

azure-pipelines-official.yml
MichaelSimons: They can be useful at times in the case of build issues w/tests.  The part that made we ask is that binlogs were enabled for linux but not mac.

mmitche (approved):

eng/common-variables.yml
ellahathaway: Fixed in https://github.com/dotnet/install-scripts/commit/167f6a34fd04e9843afb2308d41f668b657e2cf1

azure-pipelines-official.yml
ellahathaway: Fixed in https://github.com/dotnet/install-scripts/commit/fe4f405ded6f2e0ba97ed348640a2f40ff5643e5

MichaelSimons (approved):

azure-pipelines-PR.yml
ellahathaway: Fixed in https://github.com/dotnet/install-scripts/pull/717/commits/01b5610896148ea1fe50245ec8d88dcdde1eb24b

